PR #21, opened 2026-02-18
fix(ingestion): byteOffset uses actual bytes read, atomic state save (v0.5.2b)

eja-assistant:
## Summary

Two targeted reliability fixes for the file watcher.

## Fix 1: byteOffset race condition (CRITICAL)

After processing a file cycle, the watcher was saving `stat.size` (from the outer stat at the top of the function) as the new byteOffset. `readFileFn` does its own inner stat on the file handle — if the file grows between those two calls (common during live AI sessions), `readFileFn` returns more bytes than the outer stat predicted, but we saved the stale outer `stat.size`. This caused double-processing on the next cycle, leading to duplicate entries in the DB.

**Fix:** `byteOffset: byteOffset + newBytes.byteLength` — save exactly how many bytes were consumed.

## Fix 2: Atomic watch state write

`saveWatchState` was writing directly to the state file with `fs.writeFile`. A crash mid-write corrupts the file, causing the watcher to lose all byte offsets on restart and re-process everything from 0.

**Fix:** Write to a `.tmp` file first, then `fs.rename` atomically over the target. On POSIX the rename is atomic — either the old file survives intact or the new file replaces it cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed file watching to prevent duplicate event processing when watched files grow during read operations.
  * Improved watch state persistence with atomic writes to prevent corruption during unexpected process termination.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
   - Session-start permanent window widened to 30 days (temporary remains shorter)
   - Dynamic budget allocation based on available categories
   - Recency tiebreaking within a 0.05 score dead-band applied to the recent category only
+- Watch ingestion now advances `byteOffset` by bytes actually read in each cycle, preventing duplicate processing when files grow during read.
+- Watch state saves are now atomic (temp file + rename), preventing partial-write corruption on process crashes.
 
 ## 0.5.0 (2026-02-17)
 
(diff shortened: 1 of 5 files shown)


jdvmi00 (approved):
LGTM.

PR #26, opened 2026-02-18
fix(eval): use correct recall path for non-session-start categories

eja-assistant:
## Problem

`agenr eval recall` was returning zero results for 4 of its 5 query categories. The `runFtsEval` function passed natural language strings (e.g. "decisions made recently") verbatim to SQLite FTS5 MATCH, which uses implicit AND — requiring every word to co-occur literally in one entry. Zero entries matched.

Meanwhile the real `agenr recall` command uses hybrid vector + FTS search and returns correct results.

## Fix

Replaced `runFtsEval` with the correct retrieval strategy per category:

- **recent-decisions** → SQL `WHERE type='decision'` ordered by recency (deterministic, no API call)
- **active-todos** → SQL `WHERE type='todo'` ordered by importance (deterministic, no API call)
- **preferences** → real `recall()` function with vector + FTS hybrid
- **architecture** → real `recall()` function with vector + FTS hybrid
- **session-start** → unchanged, already correct

## Result

All 5 categories now return meaningful results representative of what users actually see in production.

## Release Notes

* **New Features**
  * Enhanced entry recall with semantic search capabilities for more accurate and contextual retrieval of decisions and todos.

* **Tests**
  * Improved test coverage for multi-path recall functionality including semantic search pathways.

### DIFF
--- a/src/commands/eval.ts
+++ b/src/commands/eval.ts
@@ -4,8 +4,9 @@ import path from "node:path";
 import type { Client } from "@libsql/client";
 import { readConfig } from "../config.js";
 import { closeDb, getDb, initDb } from "../db/client.js";
-import { getTagsForEntryIds, mapStoredEntry, scoreEntry } from "../db/recall.js";
+import { getTagsForEntryIds, mapStoredEntry, recall, scoreEntry } from "../db/recall.js";
 import { sessionStartRecall } from "../db/session-start.js";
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #42, opened 2026-02-19
feat(watch): add watcher PID guard for ingest (v0.6.5)

eja-assistant:
## Summary

Prevents write conflicts between `agenr ingest` and `agenr watch` by blocking ingest when the watcher is running.

## What Changed

- `src/watch/pid.ts` (new): `writeWatcherPid`, `deleteWatcherPid`, `readWatcherPid`, `isProcessAlive`, `isWatcherRunning` - stale-PID detection via `process.kill(pid, 0)` (EPERM = alive, ESRCH = dead)
- `src/commands/watch.ts`: writes `watcher.pid` on start, deletes in `finally` block; `onShutdown` hook registered as v0.6.6 prep
- `src/commands/ingest.ts`: guard at entry point before `installSignalHandlers()`; exits 1 with clear error if watcher is running; `options.force` does NOT bypass the guard

## Polish (commit 2)
- `formatError` used consistently for PID write failure in watch.ts (was `formatWarn`)
- ingest watcher-blocked error uses `clack.log.error` with `formatError` (was raw `process.stderr.write`)
- `clack.intro()` moved before watcher guard so it fires exactly once on all paths

## Tests

668 tests passing (+19 new). Coverage includes:
- Stale PID overwrite test (would catch `flag:'wx'` regression)
- Write ordering assertion via `invocationCallOrder`
- Both normal and error-exit cleanup paths
- `options.force` does not bypass guard
- Full early-return shape verification (all stats zeroed)

## Release Notes v0.6.5

* **New Features**
  * Watcher now manages a PID file on startup and cleanup on exit for process tracking
  * Ingest command now detects and blocks execution when watcher is running to prevent conflicts
  * Added stale-PID detection for robust watcher monitoring

* **Bug Fixes**
  * Improved error message consistency for watcher-related issues with proper error-level formatting
  * Enhanced conflict prevention between ingest and watcher processes at entry point

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [0.6.5] - 2026-02-19
+
+### Added
+- feat(watch): watcher writes watcher.pid on start and deletes on exit
+- feat(ingest): ingest exits 1 with clear error if watcher is running
+- feat(watch): isWatcherRunning() helper with stale-PID detection in src/watch/pid.ts
+- feat(watch): deleteWatcherPid registered via onShutdown() as v0.6.6 graceful shutdown hook point
+
+### Fixed
+- fix(ingest): write conflicts between ingest and watcher are now blocked at the ingest entry point
+
 ## [0.6.4] - 2026-02-19
 
 ### Added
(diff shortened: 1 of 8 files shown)


jdvmi00 (approved):
LGTM - swarm reviewed, 668 passing.

jdvmi00 (approved):
LGTM - swarm reviewed, 668 passing.

PR #45, opened 2026-02-19
fix(openclaw-plugin): use api.on(before_agent_start) instead of registerHook (v0.6.8)

eja-assistant:
## Problem

The OpenClaw plugin registered the memory injection handler via `api.registerHook("agent:bootstrap", ...)` but the hook never fired.

**Root cause:** `api.registerHook` adds the handler to the internal `handlers` map inside the gateway bundle (`entry.js`, inlined code). When a new session starts, `applyBootstrapHookOverrides` fires `triggerInternalHook` from the embedded agent runner (`pi-embedded-BxoxxVJz.js`), which imports its registry from the external `registry-B3v_dMjW.js` module. These are two separate `handlers` map instances. The handler was never in the embedded agent's map. `allHandlers.length === 0` and the event returned without calling anything.

## Fix

Switch to `api.on("before_agent_start", handler)`, which uses the typed hook system stored in the global plugin registry (`Symbol.for("openclaw.pluginRegistryState")`). This IS shared across both bundles. Return `{ prependContext: markdown }` to inject content.

This is the same pattern used by the stock `memory-lancedb` plugin.

## Changes

- `src/openclaw-plugin/index.ts`: swap `registerHook` -> `api.on`, return `{ prependContext }` instead of mutating `bootstrapFiles`
- `src/openclaw-plugin/types.ts`: remove bootstrap event types, add `BeforeAgentStartEvent` / `BeforeAgentStartResult`
- `tests/openclaw-plugin/plugin.test.ts`: updated tests for new hook signature
- `package.json`: version bump to 0.6.8
- `CHANGELOG.md`: v0.6.8 entry

## Testing

- Unit tests updated and passing
- Verified against memory-lancedb reference implementation

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed openclaw-plugin handler registration so long-term memory injection initializes reliably across sessions and bundles.
* **Release**
  * Bumped package to version 0.6.8 and added corresponding changelog entry.
* **Tests**
  * Updated lifecycle tests to validate the new initialization flow and expected outcomes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# Changelog
+
 ## [0.6.8] - 2026-02-19
 
 ### Fixed
@@ -7,8 +9,6 @@
 
 ### Fixed
 - fix(openclaw-plugin): add name and description to registerHook opts to resolve OpenClaw hook registration warning
-
-# Changelog
 
 ## [0.6.6] - 2026-02-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [0.6.8] - 2026-02-19
+
+### Fixed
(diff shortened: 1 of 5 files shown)


jdvmi00 (dismissed):
LGTM.

jdvmi00 (approved):
LGTM.

PR #46, opened 2026-02-19
fix(openclaw-plugin): session dedup, ctx.sessionKey, path fix, spawn detection, AGENR.md write (v0.6.9)

eja-assistant:
## Summary

Five fixes and one new feature for the OpenClaw plugin. All changes in `src/openclaw-plugin/` only.

### Bugs Fixed

**1. sessionKey read from wrong object**
`before_agent_start` passes context (sessionKey, workspaceDir) as the second handler argument, not on the event. The plugin was reading `event.sessionKey` which was always `undefined`, so the skip-session guard never worked correctly.

**2. Fires on every turn instead of once per session**
`before_agent_start` fires before every agent run, not just once at session start. This caused agenr recall to run on every user message (~350ms latency per turn) and memory context to be prepended to every message. Fixed with a module-level `seenSessions` Set that deduplicates per session key.

**3. DEFAULT_AGENR_PATH used wrong relative path**
The old fallback path only worked on the developer machine at `~/Code/agenr-local`. Any npm-installed user got silent recall failures. Fixed to compute path relative to the module using `import.meta.url`, resolving correctly from both `src/` (tests) and `dist/` (runtime) layouts.

**4. Spawn strategy always used node for any binary**
`runRecall` always spawned `node <agenrPath>`. If `AGENR_BIN` or `agenrPath` pointed to the npm-installed shell-script binary rather than a .js file, the spawn failed silently. Fixed with `buildSpawnArgs`: detects `.js` extension and uses `process.execPath`, otherwise spawns the binary directly.

### New Feature

**5. Writes AGENR.md to workspace directory after successful recall**
After recall, the plugin writes the memory markdown to `ctx.workspaceDir/AGENR.md` fire-and-forget (does not block session start). The file is for inspection and TUI visibility. It does not auto-load into Project Context (`loadWorkspaceBootstrapFiles` uses a hardcoded list). A separate GitHub issue has been filed with the OpenClaw team requesting `extraWorkspaceFiles` config support.

### Types Updated

- `BeforeAgentStartEvent`: removed `sessionKey` (does not exist on the event object)
- `PluginHookAgentContext` added (sessionKey, workspaceDir from second handler arg)
- `PluginApi.on` updated to declare 2-arg handler signature

## Tests

25 tests, all passing. Covers session dedup, ctx.sessionKey sourcing, buildSpawnArgs for both .js and binary paths, writeAgenrMd success and silent error handling, and all existing recall/format behaviors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session tracking now prevents duplicate recall runs per session; session key is read from the runtime context.
  * Fixed CLI path resolution and improved detection of JS vs native executables when launching recall.

* **New Features**
  * Automatically writes AGENR.md into the workspace after a successful recall (fire-and-forget).
  * Recall output gains a concise summary format for easier review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [0.6.9] - 2026-02-19
+
+### Fixed
+- OpenClaw plugin: session-seen guard prevents recall firing on every turn (fires once per session)
+- OpenClaw plugin: sessionKey now read from ctx (second handler arg) instead of event
+- OpenClaw plugin: DEFAULT_AGENR_PATH uses correct 2-level relative path to dist/cli.js
+- OpenClaw plugin: spawn strategy detects .js vs executable binary
+
+### Added
+- OpenClaw plugin: writes AGENR.md to ctx.workspaceDir after successful recall (fire-and-forget)
+
 ## [0.6.8] - 2026-02-19
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.6.8",
+  "version": "0.6.9",
   "openclaw": {
     "extensions": ["dist/openclaw-plugin/index.js"]
   },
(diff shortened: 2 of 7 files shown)


jdvmi00 (approved):
LGTM

PR #49, opened 2026-02-19
fix(shutdown): graceful shutdown wake + force-exit guard (v0.6.11)

eja-assistant:
## Summary

Fixes three gaps in the graceful shutdown path:

1. **Wake on signal**: SIGINT/SIGTERM now immediately interrupt the polling sleep via `onWake()` callback. Previously the watcher could sleep up to `intervalMs` (default 5min) before noticing shutdown.

2. **Wake callback lifecycle**: `runWatcher` registers `requestWake` on entry and deregisters via `onWake(() => undefined)` in its `finally` block. Prevents stale closures after `--once` exits or repeated runs.

3. **Force-exit guard in `runWatchCommand`**: Signal-triggered exits call `runShutdownHandlers()` (LIFO). A 5s `.unref()` timer fires `exitProcessFn(1)` if cleanup stalls. Normal completion calls `exitProcessFn(0)`. Clean `--once` exits bypass shutdown handlers and call `deleteWatcherPidFn` directly.

## Tests
- 56/56 passing (17 new tests)
- Covers: wake fires on SIGTERM, deregisters on normal exit, interrupts sleep, shutdown handlers run on signal path only, clean exits skip handlers

## Checklist
- [x] Version bumped to 0.6.11
- [x] CHANGELOG updated
- [x] All tests pass
- [x] No new TS errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown responsiveness by interrupting long-poll sleeps via a wake mechanism.
  * Prevented stale shutdown handlers across repeated runs and normal exits by always registering/deregistering handlers.
  * Ensured shutdown handlers run on signal exits and added a 5s force-exit guard to avoid hangs.

* **Chores**
  * Bumped package version to 0.6.11 and added changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.6.11] - 2026-02-19
+
+### Fixed
+- fix(shutdown): SIGINT/SIGTERM now wake the watcher immediately via a shared wake callback, so long polling sleeps are interrupted without waiting for the next interval
+- fix(watch): watcher now registers and always deregisters the shutdown wake callback in `runWatcher`, preventing stale wake handlers across normal exits (`--once`) and repeated runs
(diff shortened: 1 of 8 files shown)


jdvmi00 (approved):
LGTM

PR #80, opened 2026-02-20
feat(plugin): bundled skill + openclaw init cleanup (v0.7.3)

eja-assistant:
## What

Two related improvements to the OpenClaw plugin experience.

## 1. Bundled SKILL.md

Adds `skills/SKILL.md` to the plugin package. When installed, OpenClaw automatically loads this skill, teaching the agent when and how to use `agenr_store` and `agenr_recall` as MCP tools.

- `agenr_store`: call immediately after decisions, preferences, lessons, events. No asking first.
- `agenr_recall`: call mid-session when you need context you don't have. Not at turn 1 — the plugin's auto-injection handles session-start recall.

Registered via `"skills": ["skills"]` in `openclaw.plugin.json`. Also completes the plugin's `configSchema` with the 4 fields that were wired in code since v0.7.0 but missing from the manifest: `signalMinImportance`, `signalMaxPerSignal`, `signalsEnabled`, `dbPath`.

## 2. openclaw init cleanup

`agenr init` no longer writes to `AGENTS.md` for the openclaw platform:

- **Removed**: `AGENTS.md` presence as auto-detection heuristic for openclaw (unreliable — Codex also uses `AGENTS.md`)
- **Changed**: `--platform openclaw` now returns `null` for `instructionsPath` and skips `upsertPromptBlock` entirely
- **Kept**: `writeMcpConfig` still runs — agents need the MCP server for `agenr_store`/`agenr_recall` tool calls

The OpenClaw plugin's `prependContext` injection makes the AGENTS.md write redundant. The bundled skill covers what AGENTS.md was doing.

## Changes

- `skills/SKILL.md` — new bundled skill
- `openclaw.plugin.json` — version 0.7.3, 4 config fields added, skills dir registered
- `src/commands/init.ts` — openclaw platform: null instructionsPath, skip AGENTS.md write
- `src/commands/init.test.ts` — 2 new tests replace old AGENTS.md auto-detect test
- `CHANGELOG.md`, `package.json` — v0.7.3

## Testing

`pnpm test` — 771/771 passing

## Summary by CodeRabbit

* **New Features**
  * Added "agenr" skill for storing and recalling knowledge during sessions.
  * Added signal configuration options: importance threshold, max entries per notification, and enable/disable toggle.
  * Added configurable database path for knowledge storage.

* **Changed**
  * Updated memory injection handling for improved platform compatibility.
  * Modified platform detection behavior.

PR #82, opened 2026-02-20
fix(init): use result.platform in memory injection summary line

eja-assistant:
CRabbit nit from PR #80.

Changes hardcoded `"OpenClaw plugin"` string to `${result.platform}` in `formatInitSummary` so the summary line correctly reflects whatever platform was used, not just openclaw.

Also updates the trailing copy from `(no AGENTS.md needed)` to `(no instructions file needed)` for platform-agnostic wording.


jdvmi00 (approved):
LGTM

PR #93, opened 2026-02-20
fix: extractor importance score inflation (v0.7.7)

eja-assistant:
## Summary

Fixes importance score inflation in the extraction prompt.

### Root Cause
Three compounding issues in `src/extractor.ts` SYSTEM_PROMPT:
1. Scores 8-9-10 were one undifferentiated band with no per-score definitions
2. No signal-cost framing -- prompt had zero knowledge that 8+ fires cross-session alerts
3. Few-shot examples used 8 for routine dev facts, anchoring the LLM to inflate everything above that to 9

### Changes
- Replaced the `## Importance (1-10)` section with per-score definitions (5 through 10)
- Added signal-cost filter: 'Does someone in another session need to know this RIGHT NOW?'
- Made 7 the explicit default workhorse; 8+ now rare and justified
- Added dev session observations rule: verified/tested/confirmed facts cap at 6
- Added 3 non-developer few-shot examples (health, personal, preference) to prevent domain bias for non-dev users

### Expected Outcome
- Score 9-10: rare (0-1 per long session)
- Score 8: occasional (1-2 max), only when cross-session alert is justified
- Score 7: majority of emitted entries
- Score 6: routine dev observations and verifications
- Prompt-only fix; no schema or logic changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Recalibrated importance scoring: default start raised to 7; 8+ reserved for cross-session alerts; lower scores deprioritized.
  * Caps and stricter rules to limit high-importance signals (rare 9/10; session cap on 8s).

* **Documentation**
  * Updated guidance and examples, added sample EVENT/FACT/PREFERENCE entries and explicit rule: explicit "remember" requests justify higher importance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - fix(extractor): increase MAX_PREFETCH_RESULTS from 3 to 5 and lower PREFETCH_SIMILARITY_THRESHOLD from 0.78 to 0.72
 - fix(extractor): increase PREFETCH_CANDIDATE_LIMIT from 10 to 15 for broader elaborative encoding candidates
 - fix(extractor): tighten extractor prompt to suppress near-variant entries already captured in DB
+- fix(extractor): recalibrate importance scoring anchors so routine verifications and test-pass observations default to 6-7; reserve 8+ for cross-session alert-worthy updates
 
 ### Added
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
LGTM

PR #96, opened 2026-02-20
fix(openclaw-plugin): dedupe session-start recall by sessionId, not sessionKey

eja-assistant:
## Problem

The `before_agent_start` hook deduplicates recall using `ctx.sessionKey` to
avoid re-running on every message. But `sessionKey` is a stable slot identifier
(e.g. `agent:main:tui`) -- it does NOT change across `/new` sessions within
the same gateway lifetime.

Result: after the first session, all subsequent `/new` sessions get
`hasSeenSession(sessionKey) === true` and the recall hook returns immediately,
injecting no memory context. Confirmed in gateway debug logs:

```
18:45:01.401Z  before_agent_start fires (1 handler)
18:45:01.409Z  embedded run start  ← 8ms later, hook returned early
```
No "hooks: prepended context to prompt" log line ever appears.

## Fix

Use `ctx.sessionId` (UUID, unique per session) as the dedup key instead of
`ctx.sessionKey`. OpenClaw already passes `sessionId` in the hook context at
runtime -- it just wasn't typed.

- `types.ts`: add `sessionId?: string` to `PluginHookAgentContext`
- `index.ts`: derive `dedupeKey = ctx.sessionId ?? sessionKey` in `before_agent_start`
- `before_prompt_build` (signals) intentionally unchanged -- signal cooldown/cap should persist across `/new` within the same slot

## Tests

Added 3 test cases covering the exact bug scenario:
1. First session with key+id-A: recall runs
2. Same key, new id-B (`/new`): recall runs again
3. Same key+id-B, second message: recall skipped (correct dedup)

## Related

OpenClaw SDK issue filed to add `sessionId` as typed field: https://github.com/openclaw/openclaw/issues/22126

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session deduplication logic to better handle session identifiers when available, ensuring more accurate session tracking.

* **Tests**
  * Added comprehensive test coverage for session deduplication behavior to validate correct operation across multiple session scenarios.

### DIFF
--- a/src/openclaw-plugin/index.test.ts
+++ b/src/openclaw-plugin/index.test.ts
@@ -2,9 +2,10 @@ import { EventEmitter } from "node:events";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import * as dbClient from "../db/client.js";
 import plugin from "./index.js";
+import * as pluginRecall from "./recall.js";
 import * as pluginSignals from "./signals.js";
 import { runExtractTool, runRecallTool, runRetireTool, runStoreTool } from "./tools.js";
-import type { BeforePromptBuildResult, PluginApi } from "./types.js";
+import type { BeforeAgentStartResult, BeforePromptBuildResult, PluginApi } from "./types.js";
 
 const { spawnMock } = vi.hoisted(() => ({
   spawnMock: vi.fn(),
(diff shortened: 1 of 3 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #113, opened 2026-02-21
feat(recall): add until upper-bound date filter and window-relative recency scoring

### LINKED ISSUE #111: agenr_recall: add 'until'/'before' parameter for date range queries
## Problem

`agenr_recall` supports a `since` parameter (e.g. `7d` or ISO date) to filter entries newer than a given date, but there is no way to set an upper bound. This makes it impossible to query a specific date range like "what do I know from two weeks ago?".

Additionally, the recency scoring in the recall algorithm strongly favors recent entries, meaning older entries are effectively buried even when they are within the `since` window. The DB has data going back to January 23rd but date-bounded recall is not surfacing it.

## Proposed Fix

Add an `until` (or `before`) parameter to `agenr_recall` that accepts the same format as `since` (ISO date or relative string like `14d`):

```ts
agenr_recall({
  query: "KVM mouse sharing",
  since: "2026-02-06",
  until: "2026-02-08",
})
```

When `until` is provided alongside `since`, recency scoring should be normalized relative to the queried window rather than the current time - otherwise entries in the middle of the range still get buried by the scoring curve.

## Use Case

"What did we talk about two weeks ago?" - direct DB query confirms the data is there (8,470 entries going back to 2026-01-23), but `agenr_recall` can't surface it by date range today.

## Related

- #110 - confidence language weighting for OpenClaw extractions

eja-assistant:
## Summary

Closes #111

Adds an `until` parameter to `agenr_recall` (and the MCP `agenr_recall` tool) that sets an upper date bound on results. Entries created after the ceiling are excluded. Pairs with `since` to support precise date windows.

## Behavior

- `until: "7d"` means ceiling = `now - 7 days`; entries newer than that are excluded
- `since: "14d" until: "7d"` = entries created in the 7-day window between 7 and 14 days ago
- `since > until` (inverted range) throws a descriptive error rather than silently returning empty results
- Recency decay anchors to `ceiling ?? now` so historical queries score properly without spurious freshness boosts
- `freshnessBoost` (1.5x live-query signal) uses real `now` only -- no boost for historical window queries
- Over-fetch multiplier (3x) applied when date bounds are active to compensate for vector-layer not filtering by date

## Changes

- `src/db/recall.ts`: `ceiling` param, window-relative recency anchor, 3x over-fetch, NaN `created_at` guard
- `src/utils/time.ts`: consolidate `parseSinceToIso` to `parseDateBound` (shared by `since` and `until`); improved error message
- `src/mcp/server.ts`: expose `until` on MCP `agenr_recall` tool with clear description
- `tests/db/recall.test.ts`: vector-path filter test, NaN guard test, point-window test, recency-score assertion
- `docs/README.md`: updated test count, date range recall examples, added confidence extraction and GUI console to What's Next

## CHANGELOG

```
## [0.7.14] - 2026-02-21

### Added
- recall --until: upper date bound for agenr_recall (MCP + CLI)
- Window-relative recency scoring: anchor decay to ceiling when until is set
- parseDateBound utility consolidates since/until parsing in src/utils/time.ts

### Changed
- freshnessBoost skipped for historical window queries (ceiling set, no real-now anchor)
- Over-fetch 3x multiplier when date bounds active to compensate for vector-layer pre-filter gap
- Inverted range (since > until) throws descriptive error instead of silently returning empty

### Fixed
- NaN guard on created_at in recall scoring
```

## Summary by CodeRabbit

* **New Features**
  * Added an inclusive upper-date bound (--until) for recall queries and temporal-windowed recall.

* **Bug Fixes**
  * Recency decay now anchors to the window end while freshness uses real query time.
  * Returns a clear error when since > until, excludes corrupt timestamps from date-bound filters, and applies a 3x interim over‑fetch to improve in-window coverage.
  * Centralized time parsing with a standardized invalid-date error message.

* **Documentation**
  * CLI, MCP, architecture, and README updated with --since/--until guidance and temporal-window semantics.

* **Tests**
  * New suites covering until parsing, windowed recall behavior, scoring, and edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 ### Changed
 - fix(recall): recency decay now anchors to the `until` ceiling for historical windows while freshness boost remains anchored to real query time
 - fix(recall): centralized `parseSinceToIso` in `src/utils/time.ts` and removed duplicate implementations from recall CLI and MCP server
+- fix(recall): added inverted date-range validation - recall now returns a descriptive error when `since > until` instead of returning an empty list
+- fix(recall): interim 3x candidate over-fetch under date bounds to improve in-window recall coverage until SQL-level date filtering is added
+- fix(recall): corrupt `created_at` values are now safely excluded under date-bound filters instead of leaking invalid rows into filtered recall
 
 ## [0.7.13] - 2026-02-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [0.7.14] - 2026-02-21
+
+### Added
+- feat(recall): added `until` upper date bound to recall query filtering in CLI, MCP, and DB recall paths (`since` + `until` now define an inclusive window)
+
+### Changed
+- fix(recall): recency decay now anchors to the `until` ceiling for historical windows while freshness boost remains anchored to real query time
+- fix(recall): centralized `parseSinceToIso` in `src/utils/time.ts` and removed duplicate implementations from recall CLI and MCP server
+
 ## [0.7.13] - 2026-02-21
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ agenr watch --platform openclaw --context ~/.agenr/CONTEXT.md
 
 ## How it works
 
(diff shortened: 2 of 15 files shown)


jdvmi00 (approved):
LGTM

PR #120, opened 2026-02-21
perf(ingest): two-phase extract+write pipeline with batched write queue (#107)

### LINKED ISSUE #107: perf(ingest): decouple extract and store phases to eliminate SQLite write lock contention
## Problem

Current ingest pipeline calls extract -> store sequentially per file within each concurrent worker. With --concurrency 20, all 20 workers finish API calls simultaneously and hammer the SQLite write lock. The result is a flood of 'Another agenr process is writing to the database. Writes may be delayed' and massive serialization overhead.

On a full re-ingest of ~540 session files this takes 6-7 hours even with Tier 4 API access -- the bottleneck is the write lock, not the API rate limit. The APIs could be hit much harder.

## Solution

Decouple extract and store into two separate phases:

1. **Extract phase** - all N workers extract from files in parallel, push extracted entries into an in-memory async queue (no DB writes)
2. **Write phase** - single dedicated writer drains the queue in batches, wrapping each batch in a transaction

Key implementation points:
- Backpressure: pause extractors if queue depth exceeds a threshold (prevent OOM on large ingests)
- Batch size: tune writes-per-transaction (e.g. 25-50 entries) for optimal SQLite throughput
- Error handling: failed writes should not lose extracted entries -- retry or surface clearly
- Progress display: show queue depth alongside extraction progress in the TUI
- The per-entry dedup logic in src/db/store.ts stays unchanged -- batching only affects transaction grouping

## Expected impact

Should reduce full re-ingest time from hours to minutes for users with high API tier access. The extract concurrency can also be raised significantly (--concurrency 50+) without making the write contention worse.

eja-assistant:
## Summary

Closes #107. Decouples the extraction phase from the write phase to eliminate SQLite write-lock contention during ingest.

### Problem

The previous pipeline called `storeEntries()` synchronously per chunk inside `onChunkComplete`, with `workerCount` hardcoded to 1 to avoid cross-file lock contention. Full re-ingests of ~540 sessions took 6-7 hours even with Tier 4 API access.

### Solution

All extraction workers push extracted entries into an in-memory async write queue. A single dedicated writer drains the queue in batches. Extraction runs at full parallelism with zero write contention.

### Changes

**New: `src/ingest/write-queue.ts`**
- `WriteQueue` class with `push()`, `cancel()`, `runExclusive()`, `drain()`, `destroy()`
- Single background writer loop draining in batches grouped by file key
- Backpressure via high-watermark (default 500 entries)
- Configurable retry (`retryOnFailure` option, default true - disabled when `--no-retry` is passed)
- Shutdown integration via `isShutdownRequested` callback
- Fully injectable via `IngestCommandDeps` for unit tests

**Modified: `src/commands/ingest.ts`**
- `workerCount` now uses `--workers` option (default 10) instead of hardcoded 1
- `withDbLock`/`dbChain` removed; replaced by write queue serialization
- `insertIngestLogForFile` and `cleanupFailedFileIngest` routed through `queue.runExclusive()`
- `embeddingApiKey` resolved eagerly before queue construction
- Progress display shows queue depth: `(queue: N)`
- `deduplicateEntriesFn` still called before `queue.push()` (in-memory, belongs with producer)

### Test coverage

- 14 `write-queue.test.ts` cases covering drain-on-empty, push-after-destroy, destroy-before-drain, concurrent workers, backpressure, retry, permanent failure, stats attribution, cancel, runExclusive ordering, shutdown signal, pendingCount
- 6 new `ingest.test.ts` cases: workers wiring, mock queue injection, dry-run no-push, invalid workers values, failure path calls cancel(), noPreFetch regression, workers capping

### Behavioral notes

- `entriesStored` now counts `added + superseded` (previously only `added`)
- With retries enabled, each outer retry round makes up to 2 inner write attempts per sub-batch (2s transient-fault hedge). `--no-retry` disables both.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable file-level parallelism flag (--workers) to speed ingestion.
  * Introduced a queued, batched write pipeline to improve concurrent writes and orderly shutdown.

* **Performance**
  * Two-phase ingest reduces SQLite write-lock contention and enables higher throughput.

* **Improvements**
  * Write retry behavior with a 2s delay and optional disable flag (--no-retry); superseded entries counted with added items.

* **Tests**
  * Expanded integration and unit tests covering concurrency, retries, cancellation, and shutdown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Performance
 - perf(ingest): two-phase extract+write pipeline eliminates SQLite write-lock contention; extraction workers run in parallel while a single background writer drains entries in batched transactions (#107)
 - feat(ingest): add `--workers` flag (default 10) for file-level parallelism; previously hardcoded to 1
+- The write queue retries each write sub-batch once on transient failure (2s delay) before surfacing the error to the outer file-level retry loop. Use `--no-retry` to disable all retries including the inner write retry.
 
 ### Changed
 - ingest: `entriesStored` now counts `added + superseded` (previously only `added`); superseded entries are written before the previous entry is marked superseded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [0.7.17] - 2026-02-21
+
+### Performance
+- perf(ingest): two-phase extract+write pipeline eliminates SQLite write-lock contention; extraction workers run in parallel while a single background writer drains entries in batched transactions (#107)
+- feat(ingest): add `--workers` flag (default 10) for file-level parallelism; previously hardcoded to 1
+
+### Changed
+- ingest: `entriesStored` now counts `added + superseded` (previously only `added`); superseded entries are written before the previous entry is marked superseded
+
 ## [0.7.16] - 2026-02-21
 
 ### Fixed
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -379,6 +379,7 @@ agenr ingest [options] <paths...>
 - `--dry-run`: extract without storing.
 - `--json`: emit JSON summary.
 - `--concurrency <n>`: parallel chunk extractions (default `5`).
+- `--workers <n>`: files to process in parallel (default `10`). Total concurrent LLM chunk calls can reach `workers x concurrency`.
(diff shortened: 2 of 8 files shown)


jdvmi00 (approved):
LGTM

PR #126, opened 2026-02-21
fix(ingest): WriteQueue backpressure deadlock + verbose start logs (v0.7.19)

### LINKED ISSUE #124: ux(ingest): add in-progress indicator so users know ingest is running and not stuck
## Problem

During ingest with `--verbose`, there are long silent gaps between output lines when processing large files (75KB+). The chunker outputs `Filtered:` lines quickly, then goes quiet while waiting for parallel LLM extraction calls to return. Users cannot tell if the process is hung or working.

Example: after a burst of `Filtered:` lines, nothing is printed for 10-30 seconds until a file completes. With 80 workers and large sessions, this is common.

## Desired Behavior

Some form of in-progress signal so users know things are running:

**Option A: Periodic heartbeat log**
Every N seconds (e.g. 5s), print a status line like:
```
[~] Processing: 12 files in flight | 47 extracted so far | elapsed 00:02:14
```

**Option B: Live counter / spinner in terminal**
A single updating status line (overwrite in place) showing active workers and queue depth.

**Option C: Per-file "started" log entry**
Print `[X/N] <filename> -- started` when a worker picks up a file, not just when it completes. Currently only completion is logged in `--verbose` mode.

Option C is the lowest-effort and probably sufficient — knowing a file started processing makes the silent gap feel intentional rather than frozen.

## Context

Reported during a 566-file ingest with `--workers 80 --concurrency 10`. Large files (75-96KB) with many tool result chunks are the main cause of long gaps.

### LINKED ISSUE #125: bug(ingest): WriteQueue deadlock at high worker counts (--workers 80)
## Problem

`agenr ingest` with `--workers 80 --concurrency 10` deadlocks after processing ~175 files. The process appears alive (low CPU, stable memory) but makes no further progress. Memory is static, no new completions are logged, and no error is emitted.

## Reproduction

```bash
agenr ingest <sessions-dir> --glob "**/*.jsonl*" --workers 80 --concurrency 10 --platform openclaw --verbose
```

Process stalls after the initial batch of files completes. No output for minutes. Verified via `ps` — RSS unchanged across multiple 5-second samples.

## Root Cause (suspected)

Backpressure deadlock:

1. 80 workers × 10 concurrency flood the WriteQueue with entries
2. Queue fills past `highWatermark` (default 500)
3. All workers block waiting for the queue to drain
4. The single background writer calls `storeEntries`, which requires embeddings API calls — these are slow under high concurrency
5. Writer cannot drain fast enough to release workers
6. All workers stay blocked indefinitely → deadlock

The WriteQueue backpressure mechanism correctly prevents queue overflow, but does not account for the case where all workers block simultaneously and the writer itself is I/O-bound on embeddings.

## Impact

Process must be killed manually. No data is lost (ingest_log tracks completed files), but the run must be restarted at lower worker counts.

## Workaround

Reduce workers to stay below the backpressure threshold:
```bash
--workers 20 --concurrency 10
```

## Possible Fixes

- **Raise or dynamically scale highWatermark** based on `--workers` count
- **Timeout on backpressure wait** — if a worker has been blocked N seconds, surface an error rather than hanging forever
- **Decouple embeddings from write path** — pre-compute embeddings in workers before pushing to WriteQueue so the writer only does DB writes (fast)
- **Limit effective parallelism at queue level** — accept at most `highWatermark / concurrency` active workers to prevent saturation

eja-assistant:
## Summary

Fixes two issues reported against large-session ingest jobs.

## Fix 1: WriteQueue backpressure deadlock (#125)

**Root cause:** The default `highWatermark` was 500. Large files (75KB+) can produce 600+ entries per batch. Combined with slow OpenAI embedding calls (~1-3s each), workers flooded the queue faster than the writer could drain it, causing `pendingEntries` to never drop below the threshold. Workers polled forever -- silent hang with no recovery path.

**Changes:**
- Raised default `highWatermark` from 500 to 2000
- Added `backpressureTimeoutMs` option (default 120s) with `Math.max(1, ...)` floor validation
- Timeout throws a clear, actionable error instead of hanging, referencing the CLI flags to tune
- New CLI flags: `--queue-high-watermark` and `--queue-backpressure-timeout-ms`

## Fix 2: Verbose mode silent gap (#124)

Long pause between chunker output and completion log with no feedback on large files. Added a `[X/N] file -- starting` log line before `processTarget` so users can see the worker is alive during extraction.

## Tests

- 9/9 write-queue tests pass (includes 2 new tests: timeout throws, backpressure clears on drain)
- 47/47 ingest tests pass (includes ordering assertion for verbose start log)
- Note: `watch-platform.test.ts` has 4 pre-existing failures (`vi.doMock`/`vi.resetModules` are Vitest APIs unsupported by Bun) -- not caused by this PR, tracked separately

## Closes

Closes #125
Closes #124

## Release Notes v0.7.19

* **Bug Fixes**
  * Fixed ingest write queue deadlock on large session files
  * Removed false-positive "another process writing" warnings during multi-worker ingest

* **New Features**
  * Added CLI options for queue tuning: `--queue-high-watermark` and `--queue-backpressure-timeout-ms`
  * Enhanced verbose mode with per-file processing status messages
  * Added `--max-retries` and `--force` options for ingest control
  * Configurable backpressure timeout with clear error messaging instead of hangs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 ### Fixed
 - fix(ingest): WriteQueue backpressure deadlock when processing large session files; raised default highWatermark from 500 to 2000 and added configurable backpressure timeout (default 120s) that surfaces a clear error instead of hanging forever (#125)
-- feat(ingest): add `--queue-high-watermark` and `--queue-timeout-ms` CLI flags for tuning write queue behavior on large ingest jobs
-- feat(ingest): verbose mode now logs `[X/N] file -- starting` before extraction begins, eliminating the silent gap during large-file processing (#124)
+
+### Added
+- feat(ingest): --queue-high-watermark and --queue-backpressure-timeout-ms CLI flags for tuning write queue behavior on large ingest jobs
+- feat(ingest): verbose mode now logs "[X/N] file -- starting" before extraction begins, eliminating the silent gap during large-file processing (#124)
 
 ## [0.7.18] - 2026-02-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.7.19] - 2026-02-21
+
+### Fixed
+- fix(ingest): WriteQueue backpressure deadlock when processing large session files; raised default highWatermark from 500 to 2000 and added configurable backpressure timeout (default 120s) that surfaces a clear error instead of hanging forever (#125)
+- feat(ingest): add `--queue-high-watermark` and `--queue-timeout-ms` CLI flags for tuning write queue behavior on large ingest jobs
+- feat(ingest): verbose mode now logs `[X/N] file -- starting` before extraction begins, eliminating the silent gap during large-file processing (#124)
+
 ## [0.7.18] - 2026-02-21
 
 ### Fixed
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -381,7 +381,7 @@ agenr ingest [options] <paths...>
 - `--concurrency <n>`: parallel chunk extractions (default `5`).
 - `--workers <n>`: files to process in parallel (default `10`). Total concurrent LLM chunk calls can reach `workers x concurrency`.
 - `--queue-high-watermark <n>`: max pending write-queue entries before workers apply backpressure (default `2000`).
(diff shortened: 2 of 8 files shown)


jdvmi00 (approved):
Swarm-reviewed. All fixes applied. Green.

PR #139, opened 2026-02-22
fix: v0.8.3 - setup key rotation, model picker, LRU cache, gpt-5-nano

### LINKED ISSUE #13: bug: setup reconfigure doesn't allow changing API keys
## Bug

`agenr setup` when a config already exists prompts "Reconfigure? yes/no". When you select yes, it only lets you change the model selection - it doesn't offer to change the API key/credentials.

## Expected

Reconfigure should let you change everything: auth method, API keys, model, and embedding settings.

## Workaround

Manually edit `~/.agenr/config.json` and update the `credentials.openaiApiKey` field.

## Reproduction

1. Run `agenr setup` with an existing config
2. Select "yes" to reconfigure
3. Only model options shown - no key prompt

### LINKED ISSUE #57: perf: EmbeddingCache needs max size + LRU eviction
## Problem

The in-memory `EmbeddingCache` has unbounded growth. During large ingests (100+ session files), the cache accumulates embeddings for every unique text chunk seen. At ~6KB per embedding vector (1024 dims × float32), 10,000 entries = ~60MB of heap. This causes:

- Steady memory growth during long ingest runs
- Potential OOM on machines with limited RAM or very large backlogs
- No benefit for entries that are unlikely to be re-queried

## Proposed Fix

Add a max size (e.g. 5,000 entries, configurable) with LRU eviction:
- When the cache is full, evict the least-recently-used entry before inserting a new one
- A simple doubly-linked list + Map implementation (O(1) get/put) is sufficient

## Notes

- This is a pure in-memory optimization; no DB schema changes needed
- The cache is most useful during dedup passes (same chunk queried multiple times in a session)
- LRU is the right policy because recent chunks are most likely to appear as near-duplicates

### LINKED ISSUE #136: bug: custom model aliases not shown in `agenr setup` model picker
## Problem

`agenr setup` does not show custom model aliases (e.g. `gpt-4.1-nano`) in the model picker, even though they are defined in `src/llm/models.ts` and work correctly at runtime.

## Root Cause

Two bugs in combination:

**Bug 1 — `src/config.ts`**: The `openai-api-key` auth method's `preferredModels` list contains `"openai/gpt-4.1-nano"` (with provider prefix). Pi's model registry stores IDs without the prefix, so the match fails.

**Bug 2 — `src/setup.ts` `modelChoicesForAuth()`**: Preferred models are validated against Pi's built-in model registry:
```ts
const allModels = getModels(provider).map((model) => model.id);
const preferred = definition.preferredModels.filter((modelId) => allModels.includes(modelId));
```
Pi's registry does not include recently-added OpenAI models (gpt-4.1-nano, gpt-4.1-mini, gpt-4.1). Any model not in Pi's list is silently filtered out before the user sees it. Agenr has its own `resolveModel()` that knows about these aliases, but it is not used for this validation.

## Fix

**src/config.ts**: Remove provider prefix from preferred model IDs:
```ts
// before
preferredModels: [gpt-4o, openai/gpt-4.1-nano, gpt-4o-mini],
// after
preferredModels: [gpt-4o, gpt-4.1-nano, gpt-4o-mini],
```

**src/setup.ts**: Validate preferred models via `resolveModel()` instead of Pi registry:
```ts
// before
const preferred = definition.preferredModels.filter((modelId) => allModels.includes(modelId));
// after
const preferred = definition.preferredModels.filter((modelId) => {
  try { resolveModel(provider, modelId); return true; }
  catch { return false; }
});
```

`resolveModel` is already imported in `setup.ts`.

## Impact

Any model added to `src/llm/models.ts` that is not in Pi's built-in registry will be silently hidden from `agenr setup`. This includes all gpt-4.1-* aliases added in v0.8.1.

eja-assistant:
## What this does

Three bug fixes targeting setup usability, model picker correctness, and memory safety, plus a new model alias.

**Fixes:**
- setup: custom model aliases (gpt-4.1-nano, gpt-4.1-mini, gpt-5-nano) now appear in the model picker for openai-api-key auth. The picker previously filtered against Pi's registry, which does not know agenr's own aliases. Now uses resolveModel() with a try/catch so only resolvable aliases are included (issue #136)
- setup: reconfigure now offers to rotate the stored API key even when the existing credential is valid. Empty input is skipped with a warning. Saved credential is noted as not re-validated (issue #13)
- setup: revert hint null-normalization regression - details?.name ?? undefined correctly normalizes null to undefined; the interim change broke that guarantee
- embeddings: EmbeddingCache is now a bounded LRU with doubly-linked list internals (O(1) get/set/evict). Default max 5000 entries. Constructor throws RangeError for maxSize < 1 (issue #57)
- models: added gpt-5-nano alias (openai/gpt-5-nano) to MODEL_ALIASES
- config: openai-api-key preferredModels updated to [gpt-4.1-nano, gpt-4.1-mini, gpt-5-nano]
- README: updated recommended model from gpt-4o-mini to gpt-4.1-nano

**New tests (29 passing):**
- 9 EmbeddingCache tests including LRU eviction order, re-set MRU promotion, RangeError on bad maxSize
- 9 setup tests including credential rotation happy path, empty input warning, decline path
- 5 model alias tests including gpt-5-nano resolution

## Issues closed
- Closes #136
- Closes #13
- Closes #57

## Summary by CodeRabbit

* **New Features**
  * Interactive credential update during setup
  * Size-bounded LRU embeddings cache with eviction and visible size metric

* **Documentation**
  * Updated recommended LLM extraction model and default model guidance
  * Changelog updated with setup and embedding improvements

* **Tests**
  * New tests for LRU cache, model alias resolution, and credential reconfiguration flows

* **Chores**
  * Version bumped to 0.8.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- setup: custom model aliases (gpt-4.1-nano, gpt-4.1-mini) now appear in
+  the model picker when using openai-api-key auth (issue #136)
+- setup: openai-api-key now prioritizes gpt-4.1-nano, gpt-4.1-mini, and
+  gpt-5-nano in preferred model selection, and adds gpt-5-nano alias
+  resolution for OpenAI model lookup
+- setup: reconfigure now offers to update stored API key even when existing
+  credential is valid (issue #13)
+- embeddings: EmbeddingCache is now bounded with LRU eviction (default
+  max 5000 entries) to prevent unbounded heap growth during large ingests
+  (issue #57)
+
 ## [0.8.2] - 2026-02-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Unreleased]
+## [0.8.3]
 
 ### Fixed
 - setup: custom model aliases (gpt-4.1-nano, gpt-4.1-mini) now appear in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixed
 - setup: custom model aliases (gpt-4.1-nano, gpt-4.1-mini) now appear in
   the model picker when using openai-api-key auth (issue #136)
+- setup: revert hint null-normalization regression (details?.name ?? undefined)
+- setup: warn user when empty credential is entered during key rotation
+- setup: note that updated credential is saved but not re-validated
 - setup: openai-api-key now prioritizes gpt-4.1-nano, gpt-4.1-mini, and
   gpt-5-nano in preferred model selection, and adds gpt-5-nano alias
   resolution for OpenAI model lookup
@@ -13,6 +16,7 @@
 - embeddings: EmbeddingCache is now bounded with LRU eviction (default
   max 5000 entries) to prevent unbounded heap growth during large ingests
   (issue #57)
+- embeddings: EmbeddingCache constructor throws RangeError for maxSize < 1
 
 ## [0.8.2] - 2026-02-22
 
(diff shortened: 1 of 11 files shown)


jdvmi00 (approved):
LGTM

PR #140, opened 2026-02-22
Fix/v0.8.3 setup cache

jdvmi00:
## What this does

Tightens the EmbeddingCache constructor guard to correctly reject NaN in addition to zero and negative values.

**Problem:** The original guard `if (maxSize < 1)` silently passes NaN because NaN comparisons always return false. A caller passing NaN would get a cache that attempts eviction on every set(), dereferencing a null tail.

**Fix:** `if (!(maxSize >= 1))` correctly catches NaN, 0, and negative values. Error message now includes the received value for easier debugging.

**Test added:** NaN input to EmbeddingCache constructor now covered alongside 0 and -1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- setup: custom model aliases (gpt-4.1-nano, gpt-4.1-mini) now appear in
+  the model picker when using openai-api-key auth (issue #136)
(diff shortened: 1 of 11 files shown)


eja-assistant (dismissed):

eja-assistant (approved):

eja-assistant:
Closing in favor of #141 - clean branch rebased on master with just the NaN guard fix.

PR #141, opened 2026-02-22
fix: tighten EmbeddingCache maxSize guard to reject NaN

eja-assistant:
## What this does

Fixes a silent bug in the EmbeddingCache constructor guard.

**Problem:** `if (maxSize < 1)` silently passes NaN because NaN comparisons always return false. A caller passing NaN gets a cache that calls evictTail() on every set(), immediately dereferencing a null tail.

**Fix:** `if (!(maxSize >= 1))` correctly rejects NaN, 0, and negative values. Error message now includes the received value for easier debugging.

**Test:** NaN added to the constructor boundary test alongside 0 and -1.

## Changes
- `src/embeddings/cache.ts` - guard + error message
- `tests/embedding-cache.test.ts` - NaN coverage

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for EmbeddingCache configuration validation to include the received value for better debugging.

* **Tests**
  * Enhanced test coverage for edge cases in configuration validation.

### DIFF
--- a/src/embeddings/cache.ts
+++ b/src/embeddings/cache.ts
@@ -11,8 +11,8 @@ export class EmbeddingCache {
   private tail: string | null = null; // least recently used key
 
   constructor(maxSize = 5000) {
-    if (maxSize < 1) {
-      throw new RangeError("EmbeddingCache maxSize must be at least 1.");
+    if (!(maxSize >= 1)) {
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
LGTM

PR #152, opened 2026-02-22
feat(ingest): --bulk mode with deferred FTS/vector rebuild and crash recovery (closes #135)

### LINKED ISSUE #135: perf: bulk ingest mode (--bulk) to defer vector index and FTS rebuild
## Problem

During large ingests, two things significantly slow down write throughput:

1. **FTS triggers** - every INSERT into `entries` fires synchronous SQLite triggers updating `entries_fts_data`, `entries_fts_docsize`, and `entries_fts_idx`. That is 3-4 extra writes per entry.
2. **libsql vector index** - `idx_entries_embedding_shadow` (HNSW/DiskANN) is updated incrementally per insert. For thousands of entries this adds significant overhead.

Standard SQLite bulk load pattern: defer both until after all writes are committed, then rebuild in one shot.

## Proposed Solution

Add a `--bulk` flag to `agenr ingest` that:

1. Drops the vector index (`idx_entries_embedding`) at the start
2. Disables FTS trigger updates (or drops/recreates the FTS table)
3. Skips vector-based dedup (falls back to FTS-only dedup or skips dedup entirely)
4. After all writes: rebuilds FTS via `INSERT INTO entries_fts(entries_fts) VALUES('rebuild')`
5. After FTS rebuild: recreates the vector index via `CREATE INDEX idx_entries_embedding ON entries (libsql_vector_idx(embedding, 'metric=cosine', 'compress_neighbors=float8', 'max_neighbors=50'))`

## Caveats

- **Dedup**: in-ingest dedup relies on `vector_top_k` which requires the index. With `--bulk`, dedup is skipped or FTS-only. Run `agenr consolidate` after to clean up duplicates.
- **Vector index rebuild time**: libsql HNSW indexing can be slow on large datasets (DiskANN algorithm). Needs benchmarking on a representative sample before relying on this overnight.
- **Crash safety**: if the process is killed mid-ingest with the index dropped, the DB is left without a vector index. Should handle this gracefully (check for missing index on startup and offer to rebuild).

## Expected Impact

Write throughput improvement should be significant - benchmarks needed, but eliminating per-insert FTS trigger overhead alone is typically a 3-5x speedup on bulk SQLite workloads.

## Related

- PR #120 (two-phase write queue) addressed write contention - this issue addresses index overhead
- After bulk ingest, always run `agenr consolidate` to handle dedup and cleanup

eja-assistant:
## Summary

Implements `--bulk` mode for `agenr ingest`, addressing the write-path bottleneck identified in issue #135. In bulk mode, FTS triggers and the vector index are dropped before the ingest run and rebuilt in a single pass after all entries are written, eliminating per-row trigger overhead and index maintenance cost.

### New: `--bulk` flag for `agenr ingest`
- Drops 3 FTS triggers and the vector index before writing
- Uses `batchSize=500` (vs 40 normal) with `BEGIN IMMEDIATE` transactions per batch
- Rebuilds FTS content and vector index in a single pass after all writes complete
- Warns the user to run `agenr consolidate` after bulk ingest (vector dedup is disabled during writes)

### New: MinHash dedup (`src/db/minhash.ts`)
- 128-hash MinHash signatures using 5-gram shingles and FNV32 with pre-seeded arrays
- Two-layer dedup: in-memory `seenNormHashes` Set (cross-batch within a run) + `findDuplicateBulk` (exact hash + MinHash scan)
- New `norm_content_hash` and `minhash_sig` columns added via schema migration with backfill

### Crash Recovery
- `_meta` flag (`bulk_ingest_state`) set before teardown, cleared only after REINDEX succeeds
- `checkAndRecoverBulkIngest()` detects interrupted bulk ingest on next startup: rebuilds missing FTS triggers and/or vector index, runs `PRAGMA integrity_check`, clears meta
- Recovery gated behind `opts.checkBulkRecovery=true` to avoid slowing normal `initDb`

## Swarm Review Fixes Applied

This branch was reviewed by an algo + test + quality swarm before PR. The following issues were caught and fixed:

- **P1**: `seenNormHashes` was updated inside the transaction before `COMMIT`, causing rollback to poison the Set and silently drop entries on retry. Fixed: local `committedHashes` merged to outer Set only after `COMMIT`.
- **P1**: `bufferToMinhashSig` threw unhandled `RangeError` on any malformed blob (wrong byte length). Fixed: byte-length guard before conversion.
- **P2**: `rebuildVectorIndex` DROP+CREATE fallback was not atomic. Fixed: wrapped in `BEGIN IMMEDIATE`.
- **P2**: Backfill ran unconditionally on all ingests, not just `--bulk`. Fixed: gated on `bulkMode`.
- **P2**: Backfill cap (5000 rows) hit silently. Fixed: warns to stderr when cap is reached.
- **P2**: Short-text MinHash fallback used raw `text` instead of normalized `chars`. Fixed.
- **P2**: `getBulkIngestMeta` silently swallowed JSON parse errors. Fixed: logs warning to stderr.

## Testing

- 953 tests pass (up from 948 on base)
- New test file: `tests/db/bulk-ingest.test.ts` (353 lines)
- Covers: MinHash unit tests, dedup (hash + minhash), crash recovery (triggers missing, index missing, both missing), backfill with existing entries, cross-batch dedup, dry-run+bulk interaction, post-rebuild state verification

## Post-Merge

After merging, run `agenr consolidate` with lower thresholds to catch near-duplicates that MinHash missed:

```
agenr consolidate --sim-threshold 0.76 --cross-type-threshold 0.82 --neighbor-limit 50
```

## Summary by CodeRabbit

* **New Features**
  * Added bulk ingest mode (--bulk flag) for optimized large-scale ingestion with deferred FTS and vector index rebuilding
  * Implemented MinHash-based duplicate detection for improved dedup accuracy during bulk operations
  * Added automatic crash recovery for interrupted bulk ingests with integrity verification

* **Bug Fixes**
  * Fixed multiple bulk-ingest related issues including transaction handling, error recovery, and metadata persistence

* **Documentation**
  * Updated architecture and CLI documentation to describe bulk ingest mode, dedup behavior, and recovery procedures


jdvmi00:
@coderabbitai help

jdvmi00:
@coderabbitai full review

eja-assistant (commented):
All three findings addressed in commits e67e879, 12eff39, and e118731:

- **Fix 1 (Major)**: Teardown ordering corrected -- `setBulkIngestMeta("writing")` now runs before `dropFtsTriggersAndIndex`, wrapped in try/catch that calls `clearBulkIngestMeta` on failure. The crash recovery flag is now always set before any DB objects are dropped.
- **Fix 2 (Nitpick)**: Removed `Object.values(record)[0]` fallback from `countFromRow` -- direct `record[key]` lookup only, returns 0 for missing key.
- **Fix 3 (Nitpick)**: `execute` mock updated to use `InStatement` type from `@libsql/client`.

953/953 tests passing.

jdvmi00 (approved):
LGTM

Issue #163, opened 2026-02-22
bug: resolveWholeFileMode silently returns false for empty messages in auto mode


jdvmi00:
## Problem

In `resolveWholeFileMode`, when auto mode is active and `messages.length === 0`, the function silently returns `false` and falls back to chunked mode:

```ts
if (messages.length === 0) {
  return false;  // no warning, no log
}
```

This happens when `detectAdapter` returns an adapter that yields 0 messages (e.g. `textAdapter` for misidentified files). The user gets no indication that whole-file was skipped for a file.

## Impact

Combined with the `detectAdapter` extension bug for `.deleted` files, every deleted session is silently processed in chunked text mode with no warning. Even after the adapter bug is fixed, empty-message files should emit a diagnostic rather than fail silently.

## Fix

Emit an `onVerbose` warning when messages is empty and whole-file auto mode would have fired, so users know the file was processed as plain text chunks rather than whole-file JSONL.

eja-assistant:
Closed by PR #165 (v0.8.7). Fixed by emitting an onVerbose warning before falling back in src/ingest/whole-file.ts.

Issue #162, opened 2026-02-22
bug: unknown model context window warning never fires in ingest due to hardcoded verbose:false


jdvmi00:
## Problem

`getContextWindowTokens` emits a warning when the model is unknown:

```ts
if (contextWindow === void 0 && verbose) {
  console.warn(`[whole-file] Unknown model "${modelId}": context window unknown, falling back to chunked mode.`);
}
```

This warning is gated on `verbose`. However, in the ingest flow `verbose` is hardcoded to `false` (see related bug). As a result, if a user specifies an unrecognized model, ingest silently falls back to chunked mode with zero indication.

## Impact

If a model is not in `MODEL_CONTEXT_WINDOWS`, whole-file auto-detect silently disables itself. The user has no way to know this happened even with `--verbose`.

## Fix

Depends on fixing the hardcoded `verbose: false` in the ingest extraction call. Once that is fixed, this warning will fire correctly.

eja-assistant:
Closed by PR #165 (v0.8.7). Resolved automatically when #161 was fixed - the unknown-model warning was gated on verbose which is now correctly propagated.

Issue #161, opened 2026-02-22
bug: verbose hardcoded to false in ingest extractKnowledgeFromChunksFn call


jdvmi00:
## Problem

In the ingest command, the call to `extractKnowledgeFromChunksFn` hardcodes `verbose: false` regardless of whether `--verbose` was passed:

```ts
const extracted = await resolvedDeps.extractKnowledgeFromChunksFn({
  verbose: false,  // hardcoded - never reflects --verbose flag
  wholeFile: resolvedWholeFile,
  ...
});
```

All `[whole-file]` log messages inside `resolveWholeFileMode` and `extractWholeFileChunkWithRetry` are gated on `params.verbose`. Since it is always `false`, users get zero confirmation that whole-file mode fired, even when running with `--verbose`.

## Impact

Users cannot tell from output whether whole-file mode is active or whether ingest silently fell back to chunked mode.

## Fix

Pass the actual verbose flag through instead of hardcoding false.

eja-assistant:
Closed by PR #165 (v0.8.7). Fixed by passing the resolved verbose boolean through in src/commands/ingest.ts.

PR #165, opened 2026-02-22
fix(ingest): whole-file silent fallbacks and .deleted adapter detection

### LINKED ISSUE #160: bug: detectAdapter falls back to textAdapter for .jsonl.deleted.TIMESTAMP files
## Problem

OpenClaw renames deleted session files by appending `.deleted.<ISO-timestamp>` to the filename:

```
34da25ee.jsonl.deleted.2026-02-22T17-00-14.068Z
```

`detectAdapter` calls `path.extname(filePath)` which returns `.068Z` for these files. No case matches, so it falls through to `textAdapter`. The text adapter returns 0 messages, which means whole-file mode never triggers and extraction quality is degraded.

## Impact

Users running `agenr ingest --glob "**/*.jsonl*"` to pick up deleted sessions get silently downgraded to chunked text extraction instead of whole-file JSONL extraction. No warning is emitted.

## Fix

Strip the `.deleted.<suffix>` portion before calling `path.extname()` in `src/adapters/registry.ts`:

```ts
const strippedPath = filePath.replace(/\.deleted\.[^/\\]+$/, "");
const ext = path.extname(strippedPath).toLowerCase();
```

## File

`src/adapters/registry.ts`

eja-assistant:
## Summary

Fixes four bugs that caused silent fallbacks from whole-file extraction mode to chunked text mode with no user-visible indication. Also fixes adapter detection for OpenClaw deleted session files.

Closes #160, #161, #162, #163

### Bug fixes

**#160 - detectAdapter wrong extension for .deleted files**
OpenClaw renames deleted sessions by appending `.deleted.<ISO-timestamp>`. `path.extname()` returns `.068Z` for these files, silently routing them to `textAdapter` instead of the correct JSONL adapter. Fixed by stripping the suffix before extension lookup in `src/adapters/registry.ts`.

**#161 - verbose hardcoded to false in ingest extraction call**
`extractKnowledgeFromChunksFn` was called with `verbose: false` regardless of the `--verbose` flag. All `[whole-file]` log messages were gated on this value and never fired. Fixed by passing the resolved `verbose` boolean through in `src/commands/ingest.ts`.

**#162 - unknown model warning never fires**
Resolved automatically by fixing #161. The `getContextWindowTokens` unknown-model warning is gated on `verbose`, which is now correctly propagated.

**#163 - resolveWholeFileMode silent false return for empty messages**
When auto mode received 0 parsed messages it silently returned false with no log. Fixed by emitting an `onVerbose` warning before falling back in `src/ingest/whole-file.ts`.

## Tests

- 2 new adapter registry tests: `.deleted` files route to correct JSONL adapter
- 1 new ingest test: `verbose: true` is passed through to extraction
- 1 new whole-file test: `onVerbose` warning emitted for empty messages in auto mode
- 957 tests passing

## Release 0.8.7

* **Bug Fixes**
  * Fixed JSONL file detection to properly strip deleted file suffixes, ensuring correct routing instead of silent fallback.
  * Verbose flag now propagates into extraction calls for whole-file diagnostics.
  * Added explicit warning when auto mode yields zero parsed messages and falls back to chunked extraction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [0.8.6]
+## [0.8.7]
 
 ### Fixed
 - fix(ingest): detect `.jsonl.deleted.<timestamp>` session files as JSONL by stripping the `.deleted.*` suffix before extension lookup, restoring OpenClaw/Codex adapter routing instead of silent text fallback (issues #160, #163)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.8.6]
+
+### Fixed
+- fix(ingest): detect `.jsonl.deleted.<timestamp>` session files as JSONL by stripping the `.deleted.*` suffix before extension lookup, restoring OpenClaw/Codex adapter routing instead of silent text fallback (issues #160, #163)
+- fix(ingest): pass the resolved ingest `verbose` flag into extraction calls so whole-file diagnostics are emitted with `--verbose`, including unknown-model context-window warnings and whole-file retry/fallback logs (issues #161, #162)
+- fix(ingest): emit an explicit `[whole-file]` verbose warning when auto mode receives zero parsed messages and falls back to chunked extraction (issue #163)
+
 ## [0.8.5]
 
 ### Added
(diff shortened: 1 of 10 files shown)


jdvmi00 (approved):
LGTM

PR #167, opened 2026-02-22
fix(ingest): model-aware output budget and remove entry hard cap

### LINKED ISSUE #166: config: MAX_ENTRIES_HARD_CAP and OUTPUT_BUDGET_TOKENS are too low and hardcoded for whole-file mode
## Problem

Two hardcoded constants limit extraction quality in whole-file mode:

**1. `MAX_ENTRIES_HARD_CAP = 100`**
When whole-file mode extracts more than 100 entries from a single session, it silently truncates to the top 100 by importance score. A long deep-work session (2-3 hours) can easily yield 200+ meaningful entries. The truncation discards real knowledge.

**2. `OUTPUT_BUDGET_TOKENS = 16384`**
The LLM output is capped at 16K tokens, which constrains how many entries the model can return in a single whole-file response. For large-context models like `gpt-4.1-nano` (supports up to 32K output), this leaves half the available output capacity on the table.

These two caps compound each other: the output token limit restricts how many entries the LLM can physically write, and then the hard cap truncates what does come through.

## Impact

- Long sessions processed in whole-file mode silently lose entries
- Large-context models are underutilized - the whole point of 1M context whole-file mode is higher extraction quality, but the output budget doesn't match
- No user-visible warning when truncation occurs at the hard cap (or if it does warn, the user has no way to change the threshold)

## Fix

1. Make both values configurable via CLI flags:
   - `--max-entries <n>` (default: keep current 100 as baseline, document recommended range)
   - `--output-tokens <n>` (default: raise to 32768 for whole-file mode, keep 16384 for chunked)

2. Or at minimum, raise the defaults:
   - `MAX_ENTRIES_HARD_CAP`: raise to 250-300
   - `OUTPUT_BUDGET_TOKENS` in whole-file mode: raise to 32768

3. Emit a clear warning (not just verbose) when the hard cap truncates entries, so users know they are losing data and can act on it.

## Notes

- The output token increase only makes sense for whole-file mode - chunked mode does not need it
- Any raise in `MAX_ENTRIES_HARD_CAP` should be paired with the output token increase or entries will be truncated before they can be written

eja-assistant:
## Summary

Closes #166

Removes the 100-entry hard cap in whole-file extraction mode and replaces it with a warning-only threshold at 500 entries. Nothing extracted by the LLM is discarded - only downstream MinHash/hash dedup decides what survives. Output token budget is now resolved per-model rather than using a hardcoded 16K default.

### Model-aware output tokens
Added `MODEL_OUTPUT_TOKENS` map with per-model max output token values. New models added to both `MODEL_CONTEXT_WINDOWS` and `MODEL_OUTPUT_TOKENS`:
- gpt-5-nano: 400K context / 131K output
- gpt-5.2-codex: 400K context / 131K output
- gpt-5.3-codex: 200K context / 100K output

`getOutputTokens(client)` resolves the correct output budget per model, falling back to 16,384 for unknowns.

### Hard cap removed
`applyEntryHardCap` (truncated to top 100 by importance) replaced with `warnOnHighEntryCount` (warns at >500, returns all entries unchanged). Dedup pipeline handles duplicates downstream.

### maxTokens plumbed through
LLM API calls in whole-file mode now pass the model-resolved `maxTokens` so the API itself uses the full model output capacity.

### usableWindowTokens updated
Takes per-model output tokens as a parameter instead of the global constant, so context-fit checks correctly account for the model's actual output budget.

## Tests

- 962 tests passing
- New: `getOutputTokens` returns correct values for all new models
- New: `usableWindowTokens` uses per-model output tokens
- New: `warnOnHighEntryCount` does not truncate, warns at threshold
- New: extractor passes model-aware `maxTokens` to whole-file LLM call

## Summary by CodeRabbit

* **Bug Fixes**
  * Per-model output token budgets applied to ingestion to improve context utilization.
  * High entry counts now emit a warning at 500 entries instead of truncating data.

* **Documentation**
  * Updated architecture docs to describe auto-detection of extraction mode, token-budgeting behavior, and the 500-entry warning policy.

* **Tests**
  * Added coverage ensuring model-specific maxTokens are propagated to whole-file extraction paths.

* **Chores**
  * Bumped package version to v0.8.8 and updated changelog.


jdvmi00 (approved):
LGTM

PR #10, opened 2026-02-18
chore: add sandbox setup script for isolated dev testing

eja-assistant:
Adds `scripts/sandbox-setup.sh` for creating an isolated dev/test environment.

- Creates separate config + DB under `sandbox/data/` (gitignored)
- Copies auth credentials from live config but redirects DB to sandbox
- Provides `sandbox-agenr` and `sandbox-reset` shell commands
- Updated CONTRIBUTING.md with sandbox docs
- Fixed local path leaks in CONTRIBUTING.md

No production code changes.

## Release Notes

* **New Features**
  * Added sandbox environment setup capability with dedicated commands for isolated testing and operations.

* **Documentation**
  * Introduced sandbox environment setup guide with initialization steps and sample commands.
  * Updated contributing documentation with improved path references.

* **Chores**
  * Added sandbox directory to version control ignore rules.

PR #175, opened 2026-02-23
fix: extraction signal quality, ingest noise, adapter detection, and daemon path

### LINKED ISSUE #168: fix(bulk): suppress pre-fetch vector index error noise during bulk ingest
## Problem

During bulk ingest, two verbose messages print once per session file, resulting in hundreds of alarming or redundant log lines:

**1. Pre-fetch vector index error (alarming)**
```
[pre-fetch] skipped: SQLITE_ERROR: vector index(search): index not found
```
The vector index is intentionally dropped at bulk start, but pre-fetch doesn't know this and logs an error for every file. End users have no way of knowing this is expected and will assume their ingest is broken.

**2. Whole-file mode parameter warning (redundant)**
```
[whole-file] interChunkDelayMs and llmConcurrency have no effect in whole-file mode
```
This prints for every single file processed in whole-file mode. In a 627-file ingest, that's 627 identical lines. It's only useful once.

## Expected behavior

1. Pre-fetch should detect bulk mode (or missing vector index) and skip silently. A single line at bulk teardown time is sufficient: `[bulk] Pre-fetch disabled: vector index unavailable during bulk ingest`
2. The whole-file parameter warning should print once at ingest start, not once per file.

## Fix

- Suppress per-file pre-fetch error when vector index is unavailable due to bulk mode
- Deduplicate the whole-file parameter warning to fire once per ingest run, not per file

### LINKED ISSUE #169: fix(ingest): .reset.TIMESTAMP files not recognized as JSONL adapter (same bug as #160)
## Problem

PR #165 (v0.8.7) fixed adapter detection for `.jsonl.deleted.TIMESTAMP` files but missed `.jsonl.reset.TIMESTAMP` files. These are produced by `agenr db reset` and have the same structure as regular session JSONL files but are not recognized by the adapter.

Current behavior:
- `.jsonl.deleted.TIMESTAMP` → correctly uses jsonl adapter (fixed in #165)
- `.jsonl.reset.TIMESTAMP` → falls through to chunked raw text mode (bug)

During a 627-file bulk ingest this surfaced as 8 files falling back to chunked mode with the message:
`[whole-file] skipping whole-file: no messages parsed from file (falling back to chunked text)`

Affected files include large sessions (85KB to 1.1MB) that produce 60-100+ raw text chunks instead of clean whole-file extraction.

## Fix

Extend the adapter detection regex from #165 to also strip `.reset.TIMESTAMP` suffix, alongside `.deleted.TIMESTAMP`. The two patterns should be treated identically — both are renamed JSONL session files that need the JSONL adapter.

## Related
- Closes #164 (adapter detection fragility)

### LINKED ISSUE #172: fix(consolidate): merge LLM returns ISO timestamp for expiry field instead of 'permanent' or 'temporary'
## Problem

During consolidate, the merge LLM occasionally returns an ISO 8601 datetime string as the `expiry` field value:

```
[merge] LLM returned invalid expiry "2026-02-21T19:28:54.672Z", falling back to "permanent"
```

The valid values are `"permanent"` or `"temporary"`. The LLM appears to be confusing the `expiry` field with `created_at` timestamps from the source entries being merged.

The fallback to `"permanent"` is correct, but the prompt should be fixed to prevent this from happening in the first place.

## Fix

1. Tighten the merge system prompt to explicitly state: `expiry must be exactly the string "permanent" or "temporary" — never a date or timestamp`
2. Optionally add an enum constraint in the JSON schema passed to the LLM so it cannot produce an invalid value

### LINKED ISSUE #174: daemon install: plist hardcodes npm global path, breaks pnpm installs
## Bug

`agenr daemon install` generates a plist with `/opt/homebrew/lib/node_modules/agenr/dist/cli.js` hardcoded as the CLI path. This is the npm global path and does not exist when agenr is installed via pnpm.

## Steps to reproduce

1. Install agenr via pnpm: `pnpm add -g agenr`
2. Run `agenr daemon install`
3. Daemon loads but immediately crashes with exit code 1

## Error

```
Error: Cannot find module '/opt/homebrew/lib/node_modules/agenr/dist/cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1447:15)
```

## Expected behavior

`agenr daemon install` should resolve the actual CLI path at install time using the running process's own path (`process.argv[1]` or `process.execPath`), not a hardcoded npm global location.

## Workaround

Manually patch the generated plist to point at the correct path:
- pnpm (macOS): `/Users/<user>/Library/pnpm/global/5/node_modules/agenr/dist/cli.js`

## Environment

- agenr: 0.8.8
- install method: pnpm global
- OS: macOS (arm64)

eja-assistant:
## Summary

Four areas of improvement across two prompt passes, each followed by a swarm review + fix cycle.

---

### Part A -- Extraction signal quality improvements

Improves the LLM extraction prompt to better capture personal user context without over-extracting.

**Changes:**
- **SKIP-first framing restored**: Personal facts are now expressed as a high-signal positive signal rather than an exception to the default SKIP rule. Prevents LLMs from weighting the carve-out too heavily.
- **FACT and EVENT type definitions** extended with personal fact guidance and habit-initiation boundary (prefer EVENT for newly-begun behaviors, FACT for established ones).
- **Durability gate** updated with a 6-month test heuristic: personal disclosures (health, diet, family, occupation, location) are durable by default; transient states (current mood, today's fatigue) are not.
- **Skip rule #8** extended: personal disclosures embedded in pleasantries should still be extracted.
- **Skip rule #10** added: transient personal states are not personal facts.
- **New few-shot examples**: RELATIONSHIP (family), PREFERENCE (diet with scoring rationale), FACT (morning routine), EVENT (job change).
- **Dash style normalized**: FACT and EVENT type definitions now use em-dash consistently with the rest of the types block.
- **Pipeline tests labeled correctly**: fakeStream tests wrapped in a describe block making clear they test pipeline wiring, not LLM semantic behavior.

### Part B -- Consolidate merge expiry fix

Closes #172

- Added prompt instruction to consolidate merge context: expiry must be exactly `"permanent"` or `"temporary"`, never a date, timestamp, or other value.
- The runtime fallback in `extractMergeResultFromToolCall` already existed; the prompt instruction is defense-in-depth.
- New test covering ISO timestamp expiry fallback with warning log assertion.

### Part C -- Adapter detection for .reset files

Closes #169

- Extended `detectAdapter` regex from `/\.deleted\.[^/\\]+$/` to `/\.(deleted|reset)\.[^/\\]+$/`.
- `.reset.TIMESTAMP` files (e.g. `session.jsonl.reset.2026-02-20T04-16-49.269Z`) now correctly route to the JSONL adapter instead of falling through to the text adapter.
- Tests: plain JSONL, .deleted, and .reset variants all resolve correctly; .reset with OpenClaw content routes to the openclaw adapter.

### Part D -- Ingest noise reduction

Closes #168

**Pre-fetch vector index noise:**
- New `isVectorIndexNotFoundMessage` helper detects the specific `SQLITE_ERROR: vector index(search): index not found` error thrown during bulk ingest when the vector index is intentionally absent.
- Silently suppressed (returns `[]` without calling `onVerbose`); all other pre-fetch errors still log normally.

**Whole-file ignored-params warning:**
- New `ExtractRunOnceFlags` interface and optional `onceFlags` parameter on `extractKnowledgeFromChunks`.
- `runIngestCommand` creates one `extractOnceFlags` object per run and passes it through, so the `[whole-file] interChunkDelayMs and llmConcurrency have no effect` warning fires once per ingest run instead of once per file.
- When `onceFlags` is not provided, old behavior is preserved (backward compatible).

### Part E -- Daemon install CLI path

Closes #174

- `resolveCliPath` correctly uses `argv[1]` resolved via the injected `argvFn` (which defaults to `process.argv` via `resolveDeps`).
- Removed erroneous `?? process.argv[1]` fallback that violated the existing DI pattern.
- New tests: argv[1]-populated path appears in plist; argv[1]-undefined falls back to cwd-relative path without using the hardcoded homebrew path.

---

## Testing

All changes include targeted tests. Full suite passes: `pnpm test`.

## Swarm Review

Each Codex pass was reviewed by a three-agent swarm (algo + test + quality) before the next prompt was written. Swarm findings drove follow-up fix commits on the same branch.

## Summary by CodeRabbit

* **New Features**
  * Enhanced handling of personal disclosures with improved durability detection using a 6-month test.
  * One-time warning deduplication across ingestion runs to reduce log noise.

* **Bug Fixes**
  * Improved error handling for missing vector indexes—errors are now suppressed when indexes are unavailable.
  * Extended file-path normalization to handle reset-suffixed files correctly.

* **Tests**
  * Added comprehensive test coverage for adapter registry, daemon CLI paths, merge expiry validation, and extraction pipeline behavior.

### DIFF
--- a/src/adapters/registry.ts
+++ b/src/adapters/registry.ts
@@ -8,7 +8,7 @@ import type { SourceAdapter } from "./types.js";
 import { vscodeCopilotAdapter } from "./vscode-copilot.js";
 
 export async function detectAdapter(filePath: string): Promise<SourceAdapter> {
-  const strippedPath = filePath.replace(/\.deleted\.[^/\\]+$/, "");
+  const strippedPath = filePath.replace(/\.(deleted|reset)\.[^/\\]+$/, "");
   const ext = path.extname(strippedPath).toLowerCase();
 
   if (ext === ".jsonl") {
--- a/src/commands/daemon.ts
+++ b/src/commands/daemon.ts
@@ -228,7 +228,7 @@ async function runLaunchctl(
 }
 
 function resolveCliPath(argv: string[]): string {
-  const cliArg = argv[1];
+  const cliArg = argv[1] ?? process.argv[1];
   if (cliArg && cliArg.trim().length > 0) {
     return path.resolve(cliArg);
   }
--- a/src/commands/daemon.ts
+++ b/src/commands/daemon.ts
@@ -228,7 +228,7 @@ async function runLaunchctl(
 }
 
 function resolveCliPath(argv: string[]): string {
-  const cliArg = argv[1] ?? process.argv[1];
+  const cliArg = argv[1];
   if (cliArg && cliArg.trim().length > 0) {
     return path.resolve(cliArg);
   }
--- a/src/commands/ingest.ts
+++ b/src/commands/ingest.ts
@@ -893,6 +893,7 @@ export async function runIngestCommand(
   let totalChunksFailed = 0;
   let filesWithChunkFailures = 0;
   const chunkStatsByFile = new Map<string, { successfulChunks: number; failedChunks: number }>();
+  const extractOnceFlags = { hasWarnedWholeFileIgnoredParams: false };
   let firstPassFailedIndexSet = new Set<number>();
   let bulkTeardownComplete = false;
   let bulkVectorRebuildDurationSeconds: number | null = null;
@@ -1062,6 +1063,7 @@ export async function runIngestCommand(
         db: options.noPreFetch ? undefined : db,
         embeddingApiKey: options.noPreFetch ? undefined : embeddingApiKey ?? undefined,
         noPreFetch: options.noPreFetch === true,
+        onceFlags: extractOnceFlags,
         onVerbose: verbose
           ? (line) => {
               clack.log.info(line, clackOutput);
--- a/src/consolidate/merge.ts
+++ b/src/consolidate/merge.ts
@@ -115,6 +115,7 @@ function buildMergeContext(cluster: Cluster): Context {
     "Only include information explicitly stated in the source entries. Do not infer or add details not present.",
     "Prefer preserving temporal changes in the merged narrative.",
     "Call merge_entries with your final merged result.",
+    'expiry must be exactly the string "permanent" or "temporary" -- never a date, timestamp, or other value.',
   ].join("\n");
 
   let contentLimit: number | undefined;
--- a/src/extractor.ts
+++ b/src/extractor.ts
@@ -22,15 +22,15 @@ import { isShutdownRequested } from "./shutdown.js";
 
 export const SYSTEM_PROMPT = `You are a selective memory extraction engine. Extract only knowledge worth remembering beyond the immediate step.
 
-Default action: SKIP. Most chunks should produce zero entries -- EXCEPT personal facts about the user (health, family, home, lifestyle, relationships), which should be captured even from casual or passing mentions.
+Default action: SKIP. Most chunks should produce zero entries. Personal disclosures (health, diet, family structure, pets, occupation, location, values, relationships) are high-signal -- do not skip them just because the mention is brief or incidental. Apply the durability gate below to all entries including personal ones.
 
 ## Types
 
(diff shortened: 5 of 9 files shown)


jdvmi00 (approved):
LGTM - swarm reviewed, all findings addressed.

Issue #178, opened 2026-02-23
feat(session-start): use previous session topic as query seed when /new prompt is low-signal


eja-assistant:
## Problem

When a user sends `/new` alone (no topic in the reset message), the new
session starts with `event.prompt` equal to `/new` or empty. The fix in
#177 correctly falls back to no-query behavior in this case -- no
regression, but no improvement either.

The user just came from a live session. That context exists. We should
use it.

## Proposed fix

Use the `before_reset` hook to stash the last user message(s) from the
ending session into the plugin's own SQLite DB, keyed by `sessionKey`.
When `before_prompt_build` fires on the new session with a low-signal
prompt, check the stash and use the previous topic as the recall query
seed.

### Flow

1. `before_reset` fires -- plugin reads the last N user messages from
   the ending session and stores a short "previous topic" string in the
   plugin DB, keyed by `sessionKey`.

2. Session resets. New `sessionId` assigned.

3. `before_prompt_build` fires for the new session. `event.prompt` is
   empty or `/new` (low signal).

4. Plugin detects low-signal prompt, checks the stash for this
   `sessionKey`, finds the previous topic, uses it as the recall query
   seed.

5. After use, the stash entry is cleared (one-shot).

### No LLM required

Just grab the last 1-2 user message texts raw. That is already
high-signal enough for a vector query. No summarization step needed.

## Open questions

1. What does the `before_reset` hook event shape actually provide? Need
   to confirm it includes session messages (or at minimum the session
   transcript path so we can read the last entry).

2. What counts as "low signal"? Probably: `event.prompt` is empty,
   whitespace only, or exactly `/new` or `/reset` after trimming.

3. Stash TTL -- the stash should expire after some reasonable window
   (e.g. 1 hour) so a stale topic from yesterday does not bleed into a
   fresh session started the next morning.

## Relationship to #177

#177 fixes the common case: interactive sessions where `event.prompt`
has real content. This issue handles the `/new` alone edge case where
`event.prompt` is low-signal. The two are complementary. This should
not block #177.

## Acceptance criteria

1. User sends `/new` alone. Next session-start recall surfaces entries
   relevant to what the previous session was doing.
2. Stash is one-shot -- does not bleed into a second subsequent reset.
3. Stash expires if not consumed within a configurable TTL (default 1h).
4. No regression for cold boots where no previous session exists.

PR #180, opened 2026-02-23
feat(plugin): seed session-start recall from inbound prompt and reset topic stash

### LINKED ISSUE #177: feat(session-start): smarter cold-start recall using prior session context as query seed
## Why This Matters

People using OpenClaw want to feel like the agent is the same agent regardless of where they are talking to it. That is the real promise of agenr -- not just "AI with memory" but **persistent agent identity across surfaces.**

Right now, Telegram-EJA and TUI-EJA know the same facts in theory, but have no sense of each other. Every session start is a cold boot. That discontinuity breaks the experience: the user walks away from Telegram, opens the TUI, and has to re-explain what they were just doing. It feels like talking to a different instance of the agent that happens to share a name.

What users actually want: pick up where they left off, on any surface, with no recap required. One agent. Everywhere.

The fix proposed here is the technical mechanism that makes that possible. When shipped, the story becomes: "Works across Telegram, Claude Code, Codex, and OpenClaw TUI -- pick up where you left off, on any surface." That is a genuinely differentiated position. No other AI memory system is solving cross-surface continuity at this level.

---

## Problem

The current session-start recall fires a generic query before the conversation exists. That means it has no topic signal -- it can only rank by recency + importance + spacing, which produces a thin, often meta-level result set that misses the most actionable context.

### What I actually experienced (written from the agent's perspective)

I am EJA, an AI assistant running on Telegram via OpenClaw. On 2026-02-22, mid-session, the gateway was restarted to load the updated agenr 0.8.6 plugin. The restart wiped my in-session context. When the new session began, session-start recall fired -- but without any topic signal, it did not surface the fact that v0.8.6 had already been published, that PR #159 had been merged, or that the mini plugin had already been updated.

As a result, I re-did work that was already done:
- Created PR #159 for the version bump (it already existed and was merged)
- Tried to publish agenr@0.8.6 to npm (already published, got a 'cannot publish over previous version' error)
- Tried to update the agenr plugin on the mini (already at 0.8.6)

Jim had to tell me 'you're a little behind, all of that has been done.' That is a real failure of the memory system. The facts WERE in the database. They just did not surface at session start.

### Second failure mode: cross-surface switching

When a user switches from an active Telegram session to the TUI, the TUI wakes up with no knowledge of the Telegram conversation. The data exists in the DB. The seed doesn't. The seed query is the only missing piece.

Note: the watcher detects file changes in ~500ms but ingests on a configurable interval (default 120s). For the seed query itself, this does not matter -- the seed reads directly from the JSONL file, bypassing the DB entirely, so there is zero lag. The DB only needs to be fresh for the recall results, and the bulk of the prior session will already be ingested from earlier cycles.

### Root cause analysis

1. The session-start query is generic. It biases toward meta-level entries about how recall itself works, not the actual project state.

2. Budget is tight: 2000 tokens, 10 entries, from a 1,354-entry database. The right facts existed but were not ranked high enough to make the cut.

3. Recency scoring helped for entries stored seconds before the restart -- but for facts stored earlier in the same session (v0.8.6 published, PR #159 merged), recency alone was not enough to compete with older high-importance entries.

4. There is no continuity signal between sessions. Each session start is a cold start with no knowledge of what the previous session was doing -- whether that was a same-surface restart or a cross-surface switch.

### Option 1: Seed session-start query from the previous session's topic

Before session-start recall runs, read the last N messages of the most recently modified session JSONL file -- regardless of which surface it came from. Use those messages to generate a topic query. This gives the recall a real signal and works across surfaces.

Example: if the last session was a Telegram conversation about agenr recall quality, the TUI session-start query seeds from that, and the right context surfaces immediately.

The seed reads directly from the JSONL file (not the DB), so ingest lag is not a factor.

The v0.8.14 archive fallback (readLatestArchivedUserMessages) already proves this pattern works -- this is the same logic applied to the prior live session rather than archived sessions.

### Option 2: Pin "in-progress" entries to session-start

Add a mechanism for the agent (or the store pipeline) to mark certain entries as in-progress or session-active. These entries are always included in session-start results regardless of score, until explicitly retired or confirmed complete.

This solves the specific failure mode: "v0.8.6 published" should have been pinned as in-progress work, guaranteed to appear at the next session start.

### Option 3: Read a workspace file as the session-start query seed

OpenClaw already loads HEARTBEAT.md into the workspace context. The session-start recall could read that file and use its Immediate Priority section as the query. This is a low-cost, high-signal approach that directly answers "what was I doing?"

This could be generalized beyond OpenClaw -- any workspace file path could serve as the query seed.

### Option 4: Last-session summary entry (automatic)

At session end (or on graceful restart), store a synthetic entry of type "event" summarizing what happened: tasks completed, decisions made, next steps. This entry gets a freshness bonus that guarantees it surfaces at the next session start.

This is the handoff note a human writes before going home.

## Recommendation

Option 1 (seed from prior session JSONL) is the most general solution, handles cross-surface switching natively, and has zero ingest lag. It is the primary fix.

Option 4 (last-session summary entry) combined with Option 2 (in-progress pinning) add a second layer of defense and are complementary.

Option 3 (workspace file seed) is the lowest-effort near-term win and could be implemented as a plugin config option today.

All four are complementary, not mutually exclusive.

## Proposed config shape (plugin)

```json
{
  "sessionStart": {
    "queryMode": "generic | prior-session | workspace-file",
    "workspaceFile": "HEARTBEAT.md",
    "priorSessionMessages": 20,
    "alwaysIncludeTags": ["in-progress"]
  }
}
```

## Acceptance criteria

1. After a gateway restart mid-session, the next session-start recall surfaces facts about what the previous session was doing -- without the agent having to ask.
2. Entries explicitly tagged in-progress are always included in session-start results.
3. No regression in recall quality for cold boots where no prior session exists.
4. After switching surfaces (e.g. active Telegram session followed by opening TUI), the TUI session-start recall seeds from the Telegram session context -- not a generic query. The user should not have to re-explain what they were working on when switching between surfaces.

## Related

- #142 (Codex session-start bootstrap) -- same cold-start problem on a different surface
- HEARTBEAT.md in OpenClaw workspace -- currently a manual workaround for exactly this problem; solving this issue would make HEARTBEAT.md less necessary

eja-assistant:
## Summary

Fixes #177 and #178.

Session-start recall previously ran with no query, so `RecallQuery.text` was always `undefined`. The embedding call was skipped entirely, and results were ranked purely by recency and category classification. This PR gives the recall a real topic signal.

### Scenario 1: Normal session (daily reset / gateway restart)

The user's first inbound message is now used as the recall query seed via `event.prompt` in the `before_prompt_build` hook. Previously this value was ignored (the parameter was named `_event`). Entries relevant to what the user actually asked now surface via vector similarity instead of recency alone.

### Scenario 2: /new with no topic

When the user sends `/new` alone, the opening prompt is low-signal. The new `before_reset` hook captures the last 3 substantive user messages from the ending session, concatenates them in chronological order, and stashes them in memory keyed by `sessionKey`. The next `before_prompt_build` finds the stash and uses it as the recall query when the prompt is thin.

A message set is only stashed if the concatenated text is >= 40 chars and >= 5 words, filtering out conversational closers like "thanks", "sounds good", or "see you on the other side".

### Scenario 3: /new with a topic

The topic portion is used as the recall query directly. The stash is consumed and discarded regardless.

## Key Design Decisions

- Stash is in-memory only (no DB changes, no schema migration)
- Stash is one-shot: consumed on first read, unconditionally, even when the prompt is high-signal
- Periodic eviction sweep runs every 5 minutes (unref'd interval) to prevent accumulation in long-running gateways
- Session-start recall timeout bumped from 5s to 10s to accommodate the embedding API call
- Query truncated at 500 chars before CLI arg push (`RECALL_QUERY_MAX_CHARS`)

## Files Changed

- `src/openclaw-plugin/session-query.ts` (new) - all session topic logic extracted here
- `src/openclaw-plugin/index.ts` - wires up the two hooks, imports from session-query
- `src/openclaw-plugin/recall.ts` - query param, timeout bump, named constant
- `src/openclaw-plugin/types.ts` - `BeforeResetEvent` type + overload for `before_reset`
- `src/openclaw-plugin/session-query.test.ts` (new) - unit tests for all helpers
- `src/openclaw-plugin/index.test.ts` - integration tests for both hooks
- `src/openclaw-plugin/recall.test.ts` (new) - spawn arg tests
- `CHANGELOG.md` - 0.8.10 entry
- `docs/OPENCLAW.md` - Session-Start Recall section + security notice for child_process scanner warning
- `package.json` - version bump to 0.8.10

## Tests

1020/1020 passing. New tests cover all branches of `isThinPrompt`, `extractLastUserText` (including null/undefined content, array content, last-3 collection), `resolveSessionQuery` (TTL boundary, one-shot consumption, high-signal path), `shouldStashTopic`, both hooks end-to-end, and spawn arg construction.

## Release Notes v0.8.10

* **New Features**
  * Session-start recall now uses vector similarity seeded from the first user message or recent substantive messages after a reset.
  * Automatic capture of the last substantive messages before a session reset to seed recall.
  * Session topic stash with hourly expiration and 5-minute cleanup sweep.

* **Changed**
  * Recall timeout increased from 5s to 10s.
  * Session topic stash requires ≥40 characters and ≥5 words.

* **Fixed**
  * Session-start recall correctly applies vector similarity when a query is present.

* **Tests**
  * Expanded test coverage for recall, stash, and session-query behaviors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [0.8.10]
+
+### Added
+- feat(plugin): session-start recall now uses the inbound user message as the recall query seed, enabling vector similarity scoring instead of pure recency ranking; entries relevant to the actual conversation topic now surface at session start (issues #177, #178)
+- feat(plugin): before_reset hook captures the last 3 substantive user messages before a /new reset and stashes them in memory; the next session-start recall uses the stash as its query seed when the opening prompt is low-signal (issues #177, #178)
+- feat(plugin): session topic stash eviction sweep runs every 5 minutes; TTL is 1 hour
+
+### Changed
+- chore(plugin): session-start recall timeout increased from 5s to 10s to accommodate the embedding API call now required when a query is present
+- chore(plugin): session topic stash requires a minimum of 40 characters and 5 words to filter out low-signal conversational closers
+- refactor(plugin): session query helpers extracted from index.ts into session-query.ts
+
+### Fixed
+- fix(plugin): session-start recall no longer skips vector similarity scoring when a query is available; previously RecallQuery.text was always undefined at session start (issue #177)
+
 ## [0.8.9]
 
 ### Added
--- a/docs/OPENCLAW.md
+++ b/docs/OPENCLAW.md
@@ -18,6 +18,12 @@ openclaw plugins install agenr
 agenr setup  # configure LLM provider + auth
 ```
 
+> **Security notice:** OpenClaw's code scanner flags a critical warning during install:
+> _"Shell command execution detected (child_process)."_
+> This is expected -- agenr shells out to its own CLI binary to run recall and store
+> operations. It does not make external network calls, read your OpenClaw credentials,
+> or exfiltrate data. The plugin source is open at https://github.com/agenr-ai/agenr.
+
 ### 2. Optional plugin config
 
 After install, agenr works with no additional config. To customize, add an agenr
@@ -51,6 +57,33 @@ agenr recall "test" --limit 3
 ```
 
 That's the wiring. Now the important part.
+
+## Session-Start Recall
+
+When a new session begins, agenr automatically seeds recall with a query so that
+relevant memory surfaces before the first response. The query source depends on
+how the session started:
+
+**Normal session (daily reset or gateway restart):**
+The user's first message is used directly as the recall query. Entries relevant
+to what the user actually asked rank higher than generic recency-based results.
+
+**After /new (no topic):**
+When the user sends /new alone, the opening prompt is low-signal. agenr captures
+the last few substantive user messages from the ending session before it resets
+and uses that content as the recall seed for the new session. This means recall
+surfaces what you were working on -- not just whatever was most recently stored.
+
+A message is only eligible for stashing if the combined recent content is at least
+40 characters and 5 words. Short conversational closers ("thanks", "sounds good",
+"see you on the other side") are filtered out.
+
+**After /new with a topic (e.g., /new let's pick up the migration work):**
+The topic portion is used as the recall query directly. The stash from the
+ending session is consumed and discarded.
+
+**Cold boot (no prior session):**
+Recall falls back to recency-based ranking with no query.
 
 ## Seed Your Agent's Memory
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.8.9",
+  "version": "0.8.10",
   "openclaw": {
     "extensions": [
       "dist/openclaw-plugin/index.js"
--- a/src/openclaw-plugin/index.test.ts
+++ b/src/openclaw-plugin/index.test.ts
@@ -1,8 +1,16 @@
 import { EventEmitter } from "node:events";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import * as dbClient from "../db/client.js";
+import { __testing } from "./index.js";
 import plugin from "./index.js";
 import * as pluginRecall from "./recall.js";
+import {
+  clearStash,
+  extractLastUserText,
+  isThinPrompt,
+  resolveSessionQuery,
+  SESSION_TOPIC_TTL_MS,
+} from "./session-query.js";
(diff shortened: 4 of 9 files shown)


jdvmi00 (approved):
LGTM - three swarm passes + CRabbit all clear. 1023/1023 tests.

PR #182, opened 2026-02-23
feat(session-query): blend stash and live prompt for session-start recall (#181)

### LINKED ISSUE #181: feat: blend stash + live prompt for session-start recall query
## Problem

When a user runs `/new` and then sends a short first message, the session-start recall query is seeded only with that first message - which is often low-signal (e.g. "did the plugin fire?", "hey", "let's keep going").

The agent does not wake up until the user sends a message - this is a hard architectural constraint. There is no hook point between `/new` and the first message. So the stash captured by `before_reset` is the **only** source of topic continuity across a `/new` boundary.

## Root Cause

`resolveSessionQuery` in `session-query.ts` currently uses the live prompt as the seed when available, and only falls back to the stash when the prompt is literally the reset command. Short-but-real first messages get used as-is, ignoring the stash entirely.

## Solution

Blend stash + live prompt in `resolveSessionQuery`:

- If stash is empty: use live prompt as-is
- If live prompt is high-signal (>= 40 chars or >= 5 words): prepend stash, append live prompt, truncate to `RECALL_QUERY_MAX_CHARS`
- If live prompt is low-signal: use stash only (short first message adds noise, not signal)

Stash goes first - it carries the richer topic context from the previous session. Live prompt appends only when it contributes signal.

## Why This Matters

This is the most common real-world pattern: user resets with `/new` then asks a short follow-up. The current code handles the edge case (`/new` alone) but misses the common case. The stash exists precisely for this scenario and should always be factored in.

## Acceptance Criteria

- [ ] `resolveSessionQuery` accepts optional stash parameter
- [ ] Blend logic: stash-first, live prompt appended when high-signal
- [ ] Low-signal threshold consistent with stash quality gate (40 chars / 5 words)
- [ ] Output capped at `RECALL_QUERY_MAX_CHARS`
- [ ] Unit tests in `session-query.test.ts` covering: no stash, stash only, high-signal blend, low-signal (stash wins)

eja-assistant:
## Summary

Fixes #181.

When a user runs `/new` and then sends a short first message (the common real-world pattern), `resolveSessionQuery` was seeding the recall query with only the low-signal live prompt - ignoring the stash entirely.

The agent does not wake up until the user sends a message (hard architectural constraint), so the stash captured by `before_reset` is the only source of topic continuity across a `/new` boundary. This PR makes it count.

## Changes

**`session-query.ts` - `resolveSessionQuery`**

| Condition | Old behavior | New behavior |
|---|---|---|
| No stash, any prompt | live prompt | live prompt (unchanged) |
| Stash + high-signal live prompt | live prompt (stash discarded) | stash + " " + live prompt (blend) |
| Stash + low-signal live prompt | live prompt (stash discarded) | stash wins |
| Thin prompt + stash | stash | stash (unchanged) |

Signal threshold reuses the existing `shouldStashTopic` gate (40 chars / 5 words). Truncation is handled downstream in `recall.ts` (`RECALL_QUERY_MAX_CHARS`).

**`index.test.ts`**

- Updated 2 existing tests that had the old behavior baked in
- Added 3 new cases:
  - Low-signal live + stash = stash wins (unit)
  - High-signal live + stash = blend (unit)
  - High-signal live + stash = blend (integration, `before_prompt_build` level)

## Test Results

1025/1025 passing.

## Release Notes

**Bug Fixes**
* Enhanced query resolution to better leverage cached content. The system now intelligently combines cached data with new prompts when appropriate, or defaults to cached content when new input is insufficient—improving query consistency and reliability.

### DIFF
--- a/src/openclaw-plugin/index.test.ts
+++ b/src/openclaw-plugin/index.test.ts
@@ -679,22 +679,19 @@ describe("before_prompt_build query seeding", () => {
     expect(resolveSessionQuery("/new", sessionKey)).toBeUndefined();
   });
 
-  it("uses high-signal prompt and consumes stash when both are present", async () => {
+  it("uses stash as query when live prompt is low-signal and stash is present", async () => {
     const runRecallMock = vi.spyOn(pluginRecall, "runRecall").mockResolvedValue(null);
     const api = makeApi({ pluginConfig: { signalsEnabled: false } });
     plugin.register(api);
     const promptHandler = getBeforePromptBuildHandler(api);
     const resetHandler = getBeforeResetHandler(api);
-    const sessionKey = "agent:main:seed-high-with-stash";
-    seedStashWithMessage(
-      resetHandler,
-      sessionKey,
-      "Please continue discussing release risks across deployment and rollback checks",
-    );
+    const sessionKey = "agent:main:seed-low-with-stash";
+    const stashedText = "Please continue discussing release risks across deployment and rollback checks";
+    seedStashWithMessage(resetHandler, sessionKey, stashedText);
 
     await promptHandler(
       { prompt: "Check release blockers" },
-      { sessionKey, sessionId: "uuid-seed-high-with-stash" },
+      { sessionKey, sessionId: "uuid-seed-low-with-stash" },
     );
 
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
LGTM

PR #187, opened 2026-02-23
fix(openclaw-plugin): webchat session-start recall fallback via archived reset sessions

### LINKED ISSUE #186: bug: before_reset hook never fires for webchat /new (sessions.reset RPC path)
## Root Cause

OpenClaw has two separate code paths for session reset:

1. **Message-based reset** (Telegram, Signal, etc.) - goes through `reply-B2UJINPw.js` -> fires `before_reset` plugin hook ✅
2. **Webchat/RPC reset** (`/new` from webchat or TUI) - goes through `sessionsHandlers["sessions.reset"]` in `gateway-cli-Cq5z-VKf.js` -> does NOT fire `before_reset` ❌

The `sessions.reset` RPC handler calls `triggerInternalHook` (OpenClaw's internal slash-command router) and then `archiveSessionTranscriptsForSession`, but never calls `hookRunner.runBeforeReset()`. So the agenr plugin's `before_reset` stash logic never runs for webchat sessions.

## Impact

Session-start recall gets no stash for webchat sessions. The query falls back to the raw live prompt (often something like "new session" or a short message), resulting in low-quality recall.

## Evidence

Debug log after `/new hey` from webchat:
```
2026-02-23T04:00:50.556Z [before_prompt_build] sessionKey="agent:main:tui" userPrompt="new session" queryText="new session"
```
No `[before_reset]` entry ever appears in the log, even though `before_reset` is correctly registered via `api.on()`.

## Fix (implemented in agenr plugin)

Since OpenClaw is unlikely to fix the upstream RPC path, the agenr plugin now falls back to reading the most recently archived session file (`*.reset.*`) when:
- It's the first `before_prompt_build` call in a new session
- The stash is empty (i.e., `before_reset` never fired)
- The live prompt is low-signal (short/generic)

New function: `readLatestArchivedMessages(sessionsDir, maxAgeMs)` in `session-query.ts`

## OpenClaw upstream

The correct fix would be for OpenClaw to call `hookRunner.runBeforeReset()` inside `sessionsHandlers["sessions.reset"]` in `gateway-cli-Cq5z-VKf.js`, same as the message-based path. Filed for visibility - not blocking our fix.

eja-assistant:
## Problem

Webchat `/new` goes through the `sessions.reset` RPC handler, which does NOT fire the `before_reset` plugin hook. This means the stash is never populated for webchat new sessions, so session-start recall has no topic seed and fires a low-quality query.

Closes #186.

## Solution

When `isFirstInSession=true`, stash is empty (`queryText === userPrompt`), and the live prompt is short (< `SESSION_TOPIC_MIN_LENGTH`), read the most recent `.reset.*` archived session JSONL file and extract the last 3 user messages to augment the recall seed.

Key design decisions:
- Augments rather than replaces: `archived + " " + userPrompt` so short-but-specific queries keep their intent
- Loops through archived files in descending mtime order, skipping files with no user messages
- 24h max age window covers daily-use gaps
- Falls back gracefully to empty on any error (non-critical path)
- Respects `agentId` from ctx with fallback to `"main"`; logs when defaulting

## Changes

- `session-query.ts`: add `readLatestArchivedUserMessages()`, `ARCHIVED_SESSION_MAX_AGE_MS` constant
- `index.ts`: archive fallback block in `before_prompt_build`, agentId resolution, debug logging
- `types.ts`: add `agentId?: string` to `PluginHookAgentContext`
- Tests: 7 new tests in `session-query.test.ts` + 5 new tests in `index.test.ts` (1045 total, all passing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced session context recovery when initiating new conversations. The system now retrieves relevant discussion history from previous sessions and incorporates it automatically, particularly for conversations beginning with brief opening prompts. This ensures important topics and context are preserved across session restarts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Changelog
 
-## [0.8.13]
+## [0.8.13] - 2026-02-23
 
 ### Fixed
-- fix(openclaw-plugin): session-start recall now falls back to reading the most recent archived OpenClaw session file (`*.reset.*`) when webchat `/new` bypasses `before_reset`; if stash-based seeding is unavailable and the opening prompt is low-signal, recall query text is built from the last 3 user messages in the archived session
+- fix(openclaw-plugin): session-start recall now falls back to reading the most recent archived OpenClaw session file (`*.reset.*`) when webchat `/new` bypasses `before_reset`. If stash-based seeding is unavailable and the opening prompt is short (< 40 characters), recall query text is built from the last 3 user messages in the archived session.
 
 ## [0.8.12]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.8.13]
+
+### Fixed
+- fix(openclaw-plugin): session-start recall now falls back to reading the most recent archived OpenClaw session file (`*.reset.*`) when webchat `/new` bypasses `before_reset`; if stash-based seeding is unavailable and the opening prompt is low-signal, recall query text is built from the last 3 user messages in the archived session
+
 ## [0.8.12]
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.8.12",
+  "version": "0.8.13",
   "openclaw": {
     "extensions": [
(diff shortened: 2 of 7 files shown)


jdvmi00 (approved):
LGTM

PR #198, opened 2026-02-23
feat(openclaw-plugin): browse cold-start recall + session handoff auto-retire (Closes #196)

### LINKED ISSUE #196: feat(openclaw-plugin): use browse mode for session-start recall when prompt is thin/low-signal
## Problem

When a new session starts with a thin or low-signal prompt (e.g. "ok cool", "hey", or after `/new`), the current session-start recall path has two failure modes:

1. **Embed path poisons itself**: `event.prompt` may contain the full agenr memory injection block. `stripPromptMetadata` strips the OpenClaw timestamp envelope but not the memory context block itself. This means `userPrompt` is thousands of chars of injected content - the archive fallback condition `userPrompt.length < SESSION_TOPIC_MIN_LENGTH` never fires, and the recall query is the injection block, not the actual user intent.

2. **Archive fallback is fragile**: Even when it does fire, the stash is in-memory and gets wiped on gateway restart (which happens on `/new` for webchat). So the archive file is the only fallback, but it only works when the prompt is genuinely short AND the metadata stripping left a clean signal.

## Solution

Wire `--browse` mode (landed in #195/v0.8.16) into the session-start path as the primary fallback when the prompt is thin or low-signal:

- When `isFirstInSession` and the resolved query is thin/low-signal, call `runRecall` with `context="browse"` and `since="1d"` instead of the embed path
- Zero OpenAI API calls, no topic guessing required
- Returns recent high-importance entries sorted by recency + importance - exactly what's needed to reconstruct "where were we?"
- Archive fallback can be simplified or removed (browse replaces it for the cold-start case)

This was explicitly called out as a use case in #195: _"Unblocks #177 (session-start cold-start - agents can now use `context='browse', since='1d'` instead of topic-guessing)"_

## Session Handoff: auto-store + auto-retire

When `/new` fires (`before_reset` hook), the plugin should automatically store a **session handoff entry** capturing where we left off:

- **type:** `event`
- **importance:** 10
- **subject:** `session handoff <timestamp>`
- **content:** current branch, open PRs, last action completed, next step queued

This entry surfaces immediately in browse at next session start (importance=10 = top of list).

To prevent it from polluting future sessions, the `before_prompt_build` hook should **auto-retire any handoff entries it surfaces via browse** (Option C) - importance=10 gets you seen once, then `agenr_retire` removes it. No schema changes needed, uses existing retire machinery.

This closes the "where did we leave off?" gap that browse alone can't answer - facts/lessons/decisions are covered by browse, but in-flight task state (current branch, next step, open PRs) is not stored anywhere queryable today.

## Acceptance Criteria

- [ ] Session starting with a thin prompt (< `SESSION_TOPIC_MIN_LENGTH` chars after metadata stripping) uses browse mode recall instead of embed recall
- [ ] Session starting with a substantive prompt still uses embed-based recall (existing behavior preserved)
- [ ] Zero OpenAI embed calls on thin-prompt session start
- [ ] Browse recall result is formatted and injected the same way as current embed recall result
- [ ] Existing `readLatestArchivedUserMessages` archive fallback logic can be removed or kept as belt-and-suspenders (decision for implementer)
- [ ] `before_reset` hook stores a session handoff entry (type=event, importance=10) capturing current branch, open PRs, last action, next step
- [ ] `before_prompt_build` hook auto-retires any handoff entries it surfaces via browse after injecting them into context
- [ ] Tests covering thin-prompt → browse path and substantive-prompt → embed path
- [ ] Tests covering handoff entry creation at `before_reset`
- [ ] Tests covering handoff entry auto-retire at `before_prompt_build`
- [ ] 1078+ tests passing

## Related

- Closes #177 (session cold-start, agents can reconstruct context without topic signal)
- Depends on #195 (browse mode, already merged)
- Related to #184 (prompt metadata stripping - root cause of failure mode 1 above)

eja-assistant:
## Summary

Closes #196.

Replaces the fragile archive-file fallback with browse-mode recall for cold-start sessions, and adds session handoff entry creation/auto-retire for continuity across `/new`.

### `src/openclaw-plugin/recall.ts`
- Added `RunRecallOptions` exported type (`context`, `since`)
- `runRecall` accepts optional 5th param; browse mode uses `--browse --since \<since\>`, omits `--budget` and query arg

### `src/openclaw-plugin/index.ts`
- **before_prompt_build**: thin-prompt cold-start (`resolveSessionQuery` returns `undefined`) now uses browse mode instead of archive file fallback; archive fallback removed
- **before_prompt_build**: after browse, auto-retires any surfaced entries whose subject starts with `session handoff` - seen once, then gone
- **before_reset**: fire-and-forget `runStoreTool` stores a type=event imp=10 handoff entry capturing last session activity; hook stays `void`
- `config` moved to outer `register()` scope; removed shadowed inner declaration
- `agentId` debug log scoped to cold-start path with corrected message text
- Debug logging added to silent `.catch()` on both retire and store

## Test Coverage
- 1083 tests passing (up from 1045 pre-#196)
- 38 new tests covering: browse cold-start path, substantive-prompt embed path, handoff store, handoff auto-retire, non-handoff entries not retired, retire failure silent, missing entry id skipped
- Archive fallback tests removed (path deleted)

## Summary by CodeRabbit

* **New Features**
  * Cold-start recall now uses browse-mode retrieval when session seed is missing.
  * Handoff storage saves a structured recent exchange summary (U:/A: turns) and a lightweight event entry to aid next-session continuity.
  * Consumed handoff entries are auto-retired at session start to avoid carryover.
  * Recall API accepts an explicit browse-mode invocation for browse-only queries.

* **Tests**
  * Expanded coverage for cold-start browse flow, handoff storage/retire behavior, exchange-summary extraction, and recall argument construction.

* **Chores**
  * Version bumped to 0.8.19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [0.8.19] - 2026-02-23
+
+### Changed
+- feat(openclaw-plugin): `before_reset` handoff store content now uses a structured recent exchange summary (`U:`/`A:` turns) instead of user-only fragments, improving cross-session handoff clarity while keeping stash-based recall seeding unchanged (issue #196)
+- feat(openclaw-plugin): added `extractLastExchangeText(messages, maxTurns?)` in `src/openclaw-plugin/session-query.ts` to collect the last 5 user-turn windows with interleaved assistant context, per-turn truncation (200 chars), and chronological `U:`/`A:` formatting
+- chore(openclaw-plugin): exported `SESSION_QUERY_LOOKBACK` from `session-query.ts` for direct test assertions
+
+### Tests
+- test(openclaw-plugin): added `extractLastExchangeText` coverage for empty input, U/A formatting, per-message truncation, 5-user-turn collection window, and no-extractable-content behavior
+- test(openclaw-plugin): updated handoff-store integration assertion to verify stored content includes exchange context prefixes (`U:`/`A:`) rather than flattened user-only text
+
 ## [0.8.18] - 2026-02-23
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [0.8.18] - 2026-02-23
+
+### Changed
+- feat(openclaw-plugin): `before_prompt_build` now uses browse-mode recall (`--browse --since 1d`) for cold session starts where no stash/query seed is available, and keeps embed/query recall for substantive or stash-seeded starts (issue #196)
+- chore(openclaw-plugin): removed archived-session fallback query synthesis from session-start recall seeding, simplifying thin-prompt startup behavior to browse vs stash/embed paths only (issue #196)
+- feat(openclaw-plugin): `before_reset` now stores a fire-and-forget `event` memory entry (`session handoff ...`) with the latest user context to support next-session handoff continuity (issue #196)
+- feat(openclaw-plugin): session-start browse results now auto-retire surfaced handoff entries after context injection to avoid repeated carryover (`reason: consumed at session start`) (issue #196)
+- feat(openclaw-plugin): `runRecall` in `src/openclaw-plugin/recall.ts` now accepts an optional context options object and maps browse context to CLI browse args while preserving existing default/session-start call behavior for unchanged callers (issue #196)
+
+### Tests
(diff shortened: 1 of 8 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #203, opened 2026-02-23
fix: isThinPrompt detects OpenClaw bare-reset boilerplate (fixes #201)

### LINKED ISSUE #201: isThinPrompt misses OpenClaw bare-reset prompt, causing wrong recall path on bare /new
### Summary

`isThinPrompt` only checks for exactly `""`, `"/new"`, and `"/reset"`. For a bare `/new` with no message, OpenClaw substitutes its own internal boilerplate as `event.prompt`:

```
"A new session was started via /new or /reset. Execute your Session Startup sequence now - read the required files before responding to the user..."
```

The agenr plugin receives this string as `event.prompt`, `isThinPrompt()` returns `false`, `queryText` gets set to that boilerplate, `isColdStart` is `false`, and the plugin does a semantic embedding search against OpenClaw's own internal reset prompt text instead of firing the browse cold-start path.

### Root Cause

```ts
// isThinPrompt - current implementation
function isThinPrompt(prompt: string): boolean {
  const trimmed = prompt.trim().toLowerCase();
  return trimmed === "" || trimmed === "/new" || trimmed === "/reset";
}
```

For a bare `/new`, OpenClaw sets `event.prompt` to:

> "A new session was started via /new or /reset. Execute your Session Startup sequence now..."

This is NOT caught by `isThinPrompt`, so:
- `queryText` = `"A new session was started via /new or /reset..."` (the boilerplate itself)
- `isColdStart = false`
- Browse path never fires
- Semantic recall runs against meaningless boilerplate

### Evidence

Confirmed by reading OpenClaw source:
```js
// pi-embedded-Vt2x_Jl3.js line 40649
const BARE_SESSION_RESET_PROMPT = "A new session was started via /new or /reset. Execute your Session Startup sequence now...";
const isBareSessionReset = isNewSession && (baseBodyTrimmedRaw.length === 0 || isBareNewOrReset);
const baseBodyFinal = isBareSessionReset ? BARE_SESSION_RESET_PROMPT : baseBody;
```

This string is passed as `event.prompt` to all `before_prompt_build` handlers.

Reproduced consistently across 3 test runs with bare `/new` on webchat.

### Fix

`isThinPrompt` needs to detect OpenClaw's bare-reset boilerplate. Two options:

**Option A - Detect known prefix (fragile if OpenClaw changes wording):**
```ts
function isThinPrompt(prompt: string): boolean {
  const trimmed = prompt.trim().toLowerCase();
  return (
    trimmed === "" ||
    trimmed === "/new" ||
    trimmed === "/reset" ||
    trimmed.startsWith("a new session was started via /new")
  );
}
```

**Option B - Detect by checking if the prompt is a known OpenClaw system injection (more robust):**
Add a `OPENCLAW_SYSTEM_PROMPT_MARKERS` list and check if the prompt matches any of them.

**Option C - Expose a flag in the hook context (requires OpenClaw change):**
OpenClaw could set `ctx.isBareReset = true` when it substitutes the bare-reset prompt, giving plugins a reliable signal. This is the cleanest solution but requires a PR upstream to openclaw/openclaw.

Option A is the quickest fix. Option C is the correct long-term solution.

### Previous Wrong Analysis

Initial investigation incorrectly blamed the slug-generator embedded run poisoning the `seenSessions` dedup map. That turned out to be wrong - the slug-generator uses `sessionKey="temp:slug-generator"` and `sessionId="slug-generator-<timestamp>"`, which are completely different from the real session's dedupeKey. No poisoning occurs. The real issue is purely in `isThinPrompt`.

eja-assistant:
## Summary

Fixes #201.

When a user sends a bare `/new` with no message, OpenClaw substitutes its internal `BARE_SESSION_RESET_PROMPT` boilerplate as `event.prompt`:

> "A new session was started via /new or /reset. Greet the user in your configured persona..."

`isThinPrompt` only checked for `""`, `"/new"`, and `"/reset"`, so the boilerplate triggered semantic recall against meaningless internal text instead of the browse cold-start path. This broke session-start recall on every bare `/new` in webchat.

## Changes

- Added `OPENCLAW_BARE_RESET_PREFIX` constant in `session-query.ts`
- Updated `isThinPrompt` to return `true` when the prompt starts with the bare-reset prefix (case-insensitive, prefix-only match for resilience to wording changes)
- Added `describe("isThinPrompt")` block with 9 test cases
- Added `describe("resolveSessionQuery - bare reset prompt")` block with 2 test cases including stash isolation via `beforeEach`/`afterAll`

## Tests

1100 passing (up from 1045).

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for session query utilities, including reset prompt detection and state management scenarios

* **Improvements**
  * Enhanced reset prompt recognition to handle additional input variations

### DIFF
--- a/src/openclaw-plugin/session-query.test.ts
+++ b/src/openclaw-plugin/session-query.test.ts
@@ -1,11 +1,15 @@
 import { mkdtemp, rm, utimes, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterAll, afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
+  clearStash,
   extractLastExchangeText,
+  isThinPrompt,
   readLatestArchivedUserMessages,
+  resolveSessionQuery,
   SESSION_QUERY_LOOKBACK,
+  stashSessionTopic,
   stripPromptMetadata,
 } from "./session-query.js";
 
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #210, opened 2026-02-24
fix(openclaw-plugin): clean up extractRecentTurns output (issue #208)

jdvmi00:
## Summary

Fixes three issues with `extractRecentTurns()` in `src/openclaw-plugin/session-query.ts` that caused noisy, truncated, and poorly formatted context in Phase 1A session injection.

## Changes

- **Strip metadata blocks** - new `stripConversationMetadata()` helper removes embedded OpenClaw `Conversation info (untrusted metadata):` JSON blocks from turn text before formatting
- **Bottom-up reading** - replaced ring buffer with collect-all + `slice(-N)` so most recent turns are always fully included when maxTurns budget is exceeded
- **Increase per-turn char limit** - `RECENT_TURN_MAX_CHARS` bumped from 150 to 300
- **Whitespace normalization** - collapse internal newlines/spaces to single space before truncating, keeps `" | "` separator clean

Also includes the `docs/OPENCLAW.md` section "Avoiding Duplicate Context Injection" (already committed to master, untouched by this branch).

## Tests

1058/1058 passing. Three new tests added: metadata strip, bottom-up selection, whitespace normalization.

Closes #208

## Summary by CodeRabbit

* **Changed**
  * Increased per-turn context limit from 150 to 300 characters.
  * Improved conversation history parsing: metadata stripping, whitespace normalization, and bottom-up selection of recent entries.
  * Enhanced session file selection to skip reset files from the current session.

* **Tests**
  * Expanded tests for conversation extraction, truncation, metadata handling, and session file behavior.

* **Chores**
  * Version bumped to 0.8.23 and release notes updated.

PR #217, opened 2026-02-24
fix(openclaw-plugin): use command hook as before_reset proxy for RPC-triggered /new

### LINKED ISSUE #216: fix: use command hook as before_reset proxy for RPC-triggered /new
## Problem

The OpenClaw `sessions.reset` RPC (used when `/new` is sent from webchat or any client calling the RPC directly) does NOT call `runBeforeReset`. This means the agenr `before_reset` hook never fires for RPC-based resets, and no handoff entry gets written to the DB before the session closes.

Investigation via gateway source (v2026.2.22-2) confirmed:
- `sessions.reset` RPC handler calls `triggerInternalHook("command", "new"...)` (Step 4) but never `runBeforeReset`
- `before_reset` is only called from the in-process auto-reply path (inline message processing)
- The `command` hook fires in BOTH paths and is awaited in the gateway path
- At command hook time, `event.context.sessionEntry.sessionFile` is still the live (unarchived) path

## Solution

Register a `command` hook handler that:
1. Filters to `action === "new" | "reset"`
2. Reads and parses the session JSONL file directly (command hook does not pre-parse messages)
3. Runs the same Phase 1 (fallback store) + Phase 2 (LLM upgrade) handoff logic
4. Uses a dedup guard (`Set<sessionId>`) to prevent double-writes when both `before_reset` and `command` fire in the auto-reply path

## Key finding

The `event.messages` array on the command hook event is an OUTPUT buffer for user replies, NOT the session transcript. The transcript must be read from `event.context.sessionEntry.sessionFile`.

## Files to change
- `src/openclaw-plugin/index.ts`
- `src/openclaw-plugin/index.test.ts`
- `CHANGELOG.md`
- `package.json` (0.8.27 -> 0.8.28)

jdvmi00:
## Problem

OpenClaw's `sessions.reset` RPC (used by webchat and any client calling the RPC directly) does NOT call `runBeforeReset`. The `before_reset` hook only fires in the in-process auto-reply path. This means no handoff entry was ever written to the DB for RPC-triggered /new commands.

Closes #210. Closes #216.

## Solution

Register a `command` hook handler that fires in both paths (RPC and auto-reply). It:
1. Filters to `action === "new" | "reset"`
2. Reads and parses the session JSONL file directly (command hook does not pre-parse messages)
3. Runs the same Phase 1 (fallback store) + Phase 2 (LLM upgrade) handoff logic via shared `runHandoffForSession()`
4. Uses a dedup guard (`Set<sessionId>` with 60s TTL eviction) to prevent double-writes when both hooks fire in the auto-reply path

## Changes

- `readAndParseSessionJsonl()` - new helper to read/parse JSONL line by line
- `runHandoffForSession()` - shared helper extracted from `before_reset` for reuse by both hooks
- `command` hook handler registered with full `[AGENR-PROBE]` debug logging at every decision point
- All `opts.logger` strings in shared helper use `opts.source` (not hardcoded `before_reset`)
- `before_reset` hook unchanged and stays in place as belt-and-suspenders for auto-reply path

## Tests

5 new tests: fires for new, skips stop, skips empty JSONL, dedup same sessionId, dedup cross-hook.
1107 tests passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed command hook handoff logic to prevent duplicate writes in RPC-triggered scenarios.
  * Improved session handling consistency across different execution paths.

* **Improvements**
  * Enhanced debugging observability for command hook operations and session file processing.

* **Tests**
  * Added new test coverage for command hook handoff behavior and deduplication scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## [0.8.28] - 2026-02-23
+
+### Fixed
+- command hook fires before_reset handoff logic for RPC-triggered /new (closes #210)
+  - before_reset hook only fires in the in-process auto-reply path; sessions.reset RPC
+    path only fires the command hook
+  - new command hook handler reads and parses the session JSONL directly, then runs
+    the same Phase 1 fallback store + Phase 2 LLM upgrade logic
+  - dedup guard (Set<sessionId>) prevents double-writes when both hooks fire in
+    auto-reply path
+
+### Added
+- [AGENR-PROBE] debug logging throughout command hook path for observability
+  (to be removed in a future cleanup pass)
+- readAndParseSessionJsonl() helper to parse JSONL session files line by line
+- runHandoffForSession() shared helper extracted from before_reset for reuse
+
+### Tests
+- 5 new tests for command hook handoff behavior in index.test.ts
+
 ## [0.8.27] - 2026-02-24
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [0.8.28] - 2026-02-23
+## [0.8.28] - 2026-02-24
 
 ### Fixed
 - command hook fires before_reset handoff logic for RPC-triggered /new (closes #210)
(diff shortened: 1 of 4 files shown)


eja-assistant (approved):

PR #222, opened 2026-02-24
fix(openclaw-plugin): handoff debug logging + fallback retirement fragility (#221)

### LINKED ISSUE #221: fix(plugin): LLM handoff debug logging + fallback retirement exact-match fragility
## Problem

Two issues discovered in `runHandoffForSession` / `summarizeSessionForHandoff` (shipped in v0.8.28-0.8.30).

### 1. No LLM visibility (debug logging gap)

`summarizeSessionForHandoff` calls the LLM silently - `verbose: false` is hardcoded and there is no debug log of:
- The transcript being sent to the LLM (length, message count)
- The LLM model + response received
- Whether the LLM returned null vs a real summary

When something goes wrong we have zero observability. Confirmed by inspecting gateway log - we see `[agenr] before_reset: calling LLM for session summary...` and then `[agenr] session_start: LLM handoff stored` with nothing in between.

### 2. Fallback retirement exact-match is fragile

`retireFallbackHandoffEntries` matches the fallback entry by exact content string equality:

```ts
if (normalizedFallbackText && entryContent !== normalizedFallbackText) {
  continue;
}
```

This fails silently if there is any whitespace/encoding difference between what was stored and what is passed in for matching. The result: both the fallback (imp:9 raw transcript) AND the LLM summary (imp:10) end up in the DB simultaneously. The fallback is never retired.

Observed in production: after a successful `[agenr] session_start: LLM handoff stored` log, the raw-transcript fallback entry remained in the DB alongside the LLM summary.

A better match strategy: match by subject + importance + handoff tag (already done), and drop the content equality check entirely. The subject is a timestamped unique key (`session handoff 2026-02-24 06:06`), so subject+imp9+tag is sufficient for safe identification.

### Fix 1 - Debug logging in summarizeSessionForHandoff

Add before the LLM call:
```
logger.debug?.(`[agenr] before_reset: sending transcript to LLM model=${resolvedModel.model} chars=${transcript.length} msgs=${currentSlice.length}+${priorSlice.length}`);
```

Add after the LLM call (on success):
```
logger.debug?.(`[agenr] before_reset: LLM summary received chars=${summaryText?.length ?? 0}`);
```

Add on null/error return paths with a reason label.

### Fix 2 - Retirement match: drop content equality check

In `retireFallbackHandoffEntries`, remove the content equality guard. Subject + importance:9 + tag:handoff is sufficient to uniquely identify the fallback entry. Content check adds fragility with zero safety benefit since subject is already a timestamped unique string.

## Version
- Affects: v0.8.28, v0.8.29, v0.8.30
- Target: v0.8.31

eja-assistant:
Closes #221

### Fix 1 - Debug logging in summarizeSessionForHandoff
- Replaces the single opaque 'calling LLM for session summary...' log with a rich pre-call log: model, transcript chars, current+prior message counts
- Adds 'LLM summary received chars=N' log on success
- Adds labeled skip-reason debug log at every null-return path (short transcript, too few messages, LLM client init failed, LLM error stop, empty summary text)

### Fix 2 - Fallback retirement drops fragile content equality check
- Removes the `entryContent !== normalizedFallbackText` guard from `retireFallbackHandoffEntries`
- Match is now subject + importance:9 + tag:handoff - sufficient since subject is a timestamped unique string
- Removes `fallbackText` param from function signature and all call sites
- Fixes production bug where fallback (raw transcript, imp:9) and LLM summary (imp:10) both remained in DB after a successful LLM run

## Tests
- 1114 passing (up from 1112, 2 new)
- New: retirement fires even when browse content differs from original fallback text
- New: debug log emitted with model + char count before LLM call

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced logging for LLM handoff interactions with detailed transcript and model metrics for improved observability
  * Resolved issue where stale fallback handoff entries would persist alongside newly generated LLM summaries

* **Tests**
  * Added test coverage for LLM handoff logging and fallback retirement behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.8.31] - 2026-02-24
+
+### Fixed
+- plugin: LLM handoff now logs transcript size, model, and summary length before/after the LLM call for observability (#221)
+- plugin: fallback handoff retirement now matches by subject+importance+tag only, dropping fragile content equality check that left stale fallback entries alongside LLM summaries (#221)
+
 ## [0.8.30] - 2026-02-24
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.8.30",
+  "version": "0.8.31",
   "openclaw": {
     "extensions": [
       "dist/openclaw-plugin/index.js"
--- a/src/openclaw-plugin/index.ts
+++ b/src/openclaw-plugin/index.ts
@@ -534,10 +534,12 @@ async function summarizeSessionForHandoff(
     let transcript = buildMergedTranscript(priorSlice, priorSurface, currentSlice, currentSurface);
     const nonHeaderLineCount = countTranscriptContentLines(transcript);
     if (nonHeaderLineCount < 3) {
+      logger.debug?.("[agenr] before_reset: skipping LLM summary - reason: short transcript");
       return null;
     }
 
     if (currentSlice.length < 5 && priorSlice.length === 0) {
+      logger.debug?.("[agenr] before_reset: skipping LLM summary - reason: too few messages");
       return null;
     }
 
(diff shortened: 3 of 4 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #224, opened 2026-02-24
fix(openclaw-plugin): raise summarizeSessionForHandoff logs from debug to console (#223)

### LINKED ISSUE #223: fix: summarizeSessionForHandoff logging uses debug level - invisible in production
## Problem

All new logging added in v0.8.31 (PR #222) inside `summarizeSessionForHandoff` uses `logger.debug?.()`. The gateway runs at `info` level or higher, so none of these lines ever appear in `gateway.err.log`.

The PROBE lines (e.g. `[AGENR-PROBE] session_start: triggering handoff`) use `console.log` directly and do appear. But after the probe fires, there is complete silence - no model, no char counts, no skip-reason - because they are all gated behind `debug`.

## Expected

After the PROBE line, you should see in `gateway.err.log`:
- Model + transcript char count + message counts before the LLM call
- Summary char count on success
- Labeled skip-reason at every null-return path (short transcript, too few messages, no apiKey, LLM error, empty summary, etc.)

## Fix

Change all `logger.debug?.()` calls inside `summarizeSessionForHandoff` to `console.log()` (consistent with the PROBE pattern already in place). No other changes needed.

## Files

- `src/openclaw-plugin/index.ts` - `summarizeSessionForHandoff` function only

eja-assistant:
Closes #223

## Changes

- Changed all `logger.debug?.()` calls inside `summarizeSessionForHandoff` to `console.log()`
- Covers all 9 log points: 5 skip-reason null paths, pre-LLM-call metadata, LLM summary received, and catch block
- Consistent with existing PROBE pattern already using `console.log`

## Tests

1115 passing (1 new: short-transcript null path in summarizeSessionForHandoff)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.8.32] - 2026-02-24
+
+### Fixed
+- summarizeSessionForHandoff: changed logger.debug to console.log for all skip-reason
+  and LLM call logging so output is visible in gateway.err.log at production log level
+  (closes #223)
+
 ## [0.8.31] - 2026-02-24
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.8.31",
+  "version": "0.8.32",
   "openclaw": {
     "extensions": [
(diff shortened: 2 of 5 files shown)


jdvmi00 (approved):
LGTM

PR #231, opened 2026-02-24
fix(openclaw-plugin): await LLM handoff upgrade in runHandoffForSession (#230)

### LINKED ISSUE #230: fix(openclaw-plugin): await LLM upgrade in runHandoffForSession - fire-and-forget loses the summary on reset
## Problem

The LLM upgrade in `runHandoffForSession` (Phase 2) is launched as `void` fire-and-forget. The `before_reset` hook IS properly awaited by the gateway - but `runHandoffForSession` returns as soon as the Phase 1 fallback is stored, before the LLM call completes. The gateway then proceeds with the reset, and the in-flight LLM promise gets dropped.

Confirmed via logs:
```
[agenr] before_reset: sending to LLM ... currentMsgs=11
[agenr] before_reset: LLM summary received chars=1699
```
...and then nothing. DB shows no LLM handoff entry. The fallback was stored and then retired at session_start, but the 1699-char LLM summary never landed.

## Root cause

```ts
// Phase 2: best-effort LLM upgrade, intentionally fire-and-forget.
if (opts.sessionFile) {
  void testingApi.summarizeSessionForHandoff(...).then(async (summary) => {
    // This runs AFTER before_reset already returned
    await runStoreTool(...);
  });
}
```

The comment even says "intentionally fire-and-forget" - but the assumption was wrong. The gateway kills the promise when the reset completes.

## Fix

Make Phase 2 fully awaited inside `runHandoffForSession`. The `before_reset` contract is `Promise<void>`, so the gateway will wait. The LLM call takes ~3 seconds - totally acceptable for a /new reset.

New flow:
1. Call `summarizeSessionForHandoff` and await it
2. If summary returned: retire fallback, store LLM entry
3. If summary failed/empty: keep fallback entry as-is (already stored in Phase 1)

Also raise the Phase 1 and Phase 2 store success/failure logs from `opts.logger?.debug?.()` to `console.log()` - same fix as #223 but for the store path.

## Files to change
- `src/openclaw-plugin/index.ts` - `runHandoffForSession` function (~line 793)

## Tests
- Add test: `before_reset` handler awaits the full LLM store before resolving (mock `summarizeSessionForHandoff` with a delay, assert store called before handler returns)
- Existing session-handoff tests should continue passing

eja-assistant:
## Problem

The LLM upgrade in Phase 2 of `runHandoffForSession` was launched as a `void` fire-and-forget. The `before_reset` hook IS awaited by the gateway, but `runHandoffForSession` returned as soon as Phase 1 (fallback store) completed. The gateway then completed the reset, killing the in-flight LLM promise.

Confirmed in logs:
```
[agenr] before_reset: LLM summary received chars=1699
```
...then nothing. The 1699-char LLM summary was generated but never stored. The next session got fallback-only quality context.

## Fix

- Replace `void .then().catch()` with `const summary = await testingApi.summarizeSessionForHandoff(...)`
- Store the LLM entry synchronously after the await resolves
- The `before_reset` contract is `Promise<void>` - the gateway waits for it, so a ~3s LLM call is fine
- Raise Phase 1 fallback store success/failure logs from `opts.logger?.debug?.()` to `console.log()` (extends #223)

## Tests

- Added test: LLM handoff store is awaited before `runHandoffForSession` resolves
- Added test: fallback-only path works correctly when LLM returns null
- 1120 tests passing (up from 1118)

Closes #230

## Release Notes v0.8.36

* **Bug Fixes**
  * Improved session handoff reliability by ensuring operations complete in the correct sequence before reset
  * Enhanced production visibility by promoting handoff operation logs from debug-only to console output

* **Chores**
  * Version bump to 0.8.36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.8.36] - 2026-02-24
+
+### Fixed
+- openclaw-plugin: await LLM upgrade in runHandoffForSession instead of fire-and-forget; the gateway awaits before_reset so the LLM call can and should block until the summary is stored (closes #230)
+- openclaw-plugin: raise Phase 1 fallback store success/failure logs from logger.debug to console.log for production visibility (extends #223)
+
 ## [0.8.34] - 2026-02-24
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.8.34",
+  "version": "0.8.36",
   "openclaw": {
     "extensions": [
       "dist/openclaw-plugin/index.js"
--- a/src/openclaw-plugin/index.test.ts
+++ b/src/openclaw-plugin/index.test.ts
@@ -1228,12 +1228,10 @@ describe("before_reset handoff store", () => {
     expect(capturedArgs[projectIdx + 1]).toBe("my-project");
   });
 
-  it("stores fallback immediately, then asynchronously upgrades with LLM summary", async () => {
-    let resolveSummary: ((value: string | null) => void) | null = null;
-    const summaryPromise = new Promise<string | null>((resolve) => {
-      resolveSummary = resolve;
-    });
-    vi.spyOn(__testing, "summarizeSessionForHandoff").mockReturnValue(summaryPromise);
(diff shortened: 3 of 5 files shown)


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #291, opened 2026-02-27
feat(recall): --around temporal targeting with gaussian recency (#189)

### LINKED ISSUE #189: feat(recall): soft recency bias parameter (--around) for approximate temporal targeting
## Problem

The `since` and `until` parameters on `agenr recall` use hard date cutoffs. If your estimate of when something happened is off by even a few days, you get zero results with no indication of why.

### Real example

When asked "what were we working on three weeks ago?", I ran:

```
agenr_recall --query "agenr project work development sessions" --since 2026-02-01 --until 2026-02-08
```

Result: 0 entries. The relevant entries existed (Jan 23-26) but were a week earlier than my estimate. The hard cutoff silently excluded them.

Only after widening the window to Jan 22 and changing query terms did relevant entries surface.

### Why hard cutoffs fail

- Human memory of "when" is approximate. "Three weeks ago" might be 18 days or 24 days.
- The caller rarely knows the exact date - they're trying to approximate.
- A miss produces zero results with no feedback, making it look like the memory doesn't exist.
- There is no way to say "bias toward this time period but don't exclude everything else."

## Proposed solution

Add a soft recency bias parameter (e.g. `--around` or `--near`) that shifts the recency scoring component toward a target date window without hard-excluding anything.

### Example behavior

```
agenr recall "what were we working on" --around 3w
```

This would:
- Boost the recency score for entries created ~3 weeks ago
- Still return entries from 2 or 4 weeks ago if they score well
- Never return zero results just because the date estimate was slightly off

### Implementation sketch

Instead of a SQL WHERE clause filtering by created_at, compute a distance score from the target date and fold it into the composite ranking score alongside vector similarity, importance, and recency. Entries near the target date get boosted; entries far from it get penalized, but nothing is hard-excluded.

### Config/API shape

```
agenr recall "<query>" --around <relative-date>     # soft center point
agenr recall "<query>" --since <date> --until <date> # existing hard cutoffs (keep for exact lookups)
```

The `--around` and `--since/--until` options would be mutually exclusive.

## Acceptance criteria

1. `agenr recall "query" --around 3w` returns relevant entries from approximately 3 weeks ago even if exact dates vary by +/- 7 days.
2. No hard zero-result failures when the target window is slightly off.
3. Existing `--since` / `--until` behavior unchanged for callers who need exact date filtering.

## Related

- #177 (session-start cold-start problem - related issue where date-based recall is needed but topic is unknown)

eja-assistant:
## Summary

Adds `--around <date>` and `--around-radius <days>` flags to recall, enabling temporal targeting that peaks recency scoring at a specific date instead of always favoring now.

## Changes

- **gaussianRecency()** - Gaussian curve centered on target date (sigma = radius/2), replacing standard recency when --around is active
- **Auto-window** - When --around is set without explicit --since/--until, auto-applies a symmetric window of +/- radius days
- **Browse mode** - Around targeting works in browse path via browseRecencyFactor() helper
- **Session-start** - Skips default 30d/7d --since windows when --around is provided
- **Plugin** - agenr_recall native tool exposes `around` and `aroundRadius` parameters
- **CLI** - `--around` and `--around-radius` wired through command options

## Testing

New tests covering gaussianRecency math (exact match, decreasing, symmetric), semantic recall with around, auto-window, explicit since/until respected, radius affects width, browse mode, session-start integration, command parsing, and plugin CLI arg wiring.

Closes #189

## Summary by CodeRabbit

* **New Features**
  * Added --around <date> and --around-radius <days> to target recall results around a specific date; browse now centers ranking on that date using a gaussian recency function and auto-applies a symmetric window when since/until are omitted.
  * agenr_recall tool exposes around and aroundRadius parameters.

* **Bug Fixes**
  * CLI now validates --around-radius (requires --around and must be positive).

* **Documentation**
  * CLI docs and README updated for around-date ranking.

* **Chores**
  * Version bumped to 0.9.6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 0.9.6 - 2026-02-27
+
+### Added
+- `--around <date>` flag for recall: shifts recency scoring to peak at a target
+  date instead of now. Entries closer to the target date rank higher regardless
+  of whether they are older or newer than it (#189).
+- `--around-radius <days>` flag: controls the window width for `--around` queries
+  (default: 14 days). Also auto-sets since/until bounds when not explicitly
+  provided.
+- `gaussianRecency` scoring function for temporal targeting.
+
 ## 0.9.5 - 2026-02-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   (default: 14 days). Also auto-sets since/until bounds when not explicitly
   provided.
 - `gaussianRecency` scoring function for temporal targeting.
+- `agenr_recall` native tool now exposes `around` and `aroundRadius` parameters
+  for temporal targeting in mid-session recall.
 
 ## 0.9.5 - 2026-02-27
 
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ agenr recall "package manager"
    tags: tooling, package-manager
 ```
 
-Recall supports date range queries (`--since 14d --until 7d`) and temporal browse mode (`--browse --since 1d`) for recency-ordered lookups without embedding API calls.
+Recall supports date range queries (`--since 14d --until 7d`), temporal browse mode (`--browse --since 1d`), and around-date targeting (`--around 2026-02-15 --around-radius 14`) to rank entries by distance from a specific date.
 
 ### Cross-session Handoff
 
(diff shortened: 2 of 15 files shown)


jdvmi00 (approved):
LGTM - around temporal targeting + CRabbit fixes all clean. 1555 tests.

PR #295, opened 2026-02-27
fix(init): resolve agenr MCP command via PATH shim

### LINKED ISSUE #294: fix: agenr init pins MCP config to exact version path instead of shim
## Problem

`resolveAgenrCommand()` in `src/commands/init.ts` uses `process.execPath` (node binary) and `process.argv[1]` (exact script path) to build the MCP server command. This hardcodes the full versioned path like:

```toml
command = "/opt/homebrew/bin/node"
args = ["/Users/jmartin/Library/pnpm/global/5/.pnpm/agenr@0.7.10_.../dist/cli.js", "mcp"]
```

When the user upgrades agenr (`pnpm add -g agenr@latest`), the MCP config still points to the old version path. The user stays pinned forever unless they manually edit the config.

## Expected

The MCP config should use the pnpm shim or resolved binary name so upgrades are automatic:

```toml
command = "agenr"
args = ["mcp"]
```

Or at minimum resolve through `which agenr` to find the shim path.

## Fix

`resolveAgenrCommand()` should try to find `agenr` on PATH first (via `which`/`where`). If found, use that path with no baseArgs. Only fall back to the process.argv approach if the binary is not on PATH.

## Impact

Every user who runs `agenr init` on any platform (Codex, Claude Code, Cursor, etc.) gets a version-pinned MCP config that breaks on upgrade.

eja-assistant:
## Summary

`resolveAgenrCommand()` now checks for the `agenr` shim/binary on PATH first, so MCP config written by `agenr init` automatically picks up new versions on upgrade. Falls back to the existing `process.execPath` approach only when the shim isn't found.

## Changes

- `src/commands/init.ts`: PATH-first resolution in `resolveAgenrCommand()`, debug logging
- `tests/commands/init.test.ts`: 59 lines of new test coverage
- `CHANGELOG.md`: v0.9.7 entry
- `package.json`: bumped to 0.9.7

## Testing

1560 tests passing. Build clean.

Closes #294

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced initialization process to dynamically resolve agenr binaries from system PATH instead of pinning to version-specific configurations, improving compatibility with software updates and supporting seamless installation upgrades.

* **Tests**
  * Added comprehensive test coverage for binary resolution during initialization, validating both successful detection and fallback mechanisms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.9.7 - 2026-02-27
+
+### Fixed
+- `agenr init` no longer pins MCP config to a version-specific path. It now
+  resolves the `agenr` shim or binary on PATH so upgrades take effect
+  automatically (#294).
+
 ## 0.9.6 - 2026-02-27
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.9.6",
+  "version": "0.9.7",
   "openclaw": {
     "extensions": [
       "dist/openclaw-plugin/index.js"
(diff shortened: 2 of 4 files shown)


jdvmi00 (approved):
LGTM

PR #304, opened 2026-02-27
feat: ingest co-recall edges, in-memory consolidate similarity, LLM dedup pre-screening (#300, #263, #264)

### LINKED ISSUE #263: perf: O(N) vector scan overhead in consolidate clustering
## Problem

During consolidate, `buildClusters` calls `findSimilar` for every candidate entry. Each call does a full vector similarity scan across all active entries (1024-dim embeddings). With the Phase 1 3x over-fetch (#249), this is amplified further.

On a 1,143-entry knowledge base (71MB), a consolidate run with ~122 estimated LLM calls takes 15+ minutes - most of that time is spent on vector scans, not LLM calls (only 26s of CPU in 11+ minutes of wall clock).

## Impact

- Consolidate is painfully slow on modest-sized knowledge bases
- Will only get worse as the brain grows
- Makes iterative consolidation impractical
- The init wizard runs consolidate post-ingest, so new users will hit this on first setup

## Possible Solutions

1. **Pre-compute similarity matrix** - build the full NxN similarity matrix once, then cluster from it instead of per-entry scans
2. **Vector index** - use an approximate nearest neighbor index (e.g. HNSW) instead of brute-force scan
3. **Batch vector queries** - group findSimilar calls and amortize the scan
4. **Cache embeddings in memory** - load all embeddings once, compute similarities in-process instead of round-tripping through SQLite

## Related

- #249 - consolidate dedup fix (introduced 3x over-fetch which amplifies this)
- #260 - consolidate benchmarking
- #262 - progress counter/spinner

### LINKED ISSUE #264: feat: LLM-assisted dedup pre-screening in consolidate clustering
## Problem

Consolidate clustering relies purely on embedding cosine similarity to decide which entries belong in the same cluster. This misses near-duplicates that are worded differently enough to fall below the similarity threshold, but express the same knowledge.

Example after running consolidate at 0.76 threshold, these all survived as separate entries:
- "Jim prefers a keto diet for weekly meal planning"
- "Jim Martin follows a ketogenic diet (keto), prefers to avoid carbohydrates"
- "Jim Martin uses keto shots and is likely following a ketogenic diet"

Embeddings see these as related but distinct. A human (or LLM) immediately recognizes them as the same fact.

## Proposed Solution

Add a two-pass approach to cluster merging:

1. **Tight threshold (e.g. 0.80+):** Auto-merge without LLM confirmation (high confidence duplicates)
2. **Loose threshold (e.g. 0.60-0.80):** LLM pre-screen - ask the model "are these truly the same knowledge, or distinct entries that happen to be related?" before merging

This could also incorporate:
- **Subject-aware clustering:** Entries with identical or near-identical subjects get clustered more aggressively regardless of embedding distance
- **Semantic dedup at store time:** Improve the existing store-time dedup to catch more duplicates before they ever land in the DB

## Why This Matters

Embedding similarity is a fundamentally lossy signal for dedup. It can't distinguish "same fact, different words" from "related but genuinely distinct facts." LLM reasoning can. Without this, consolidate will always leave residual near-duplicates that accumulate over time.

## Related

- #249 - consolidate dedup fix (Phase 3)
- #260 - consolidate benchmarking
- #263 - O(N) vector scan perf

### LINKED ISSUE #300: feat: build co-recall edges during ingest
## Problem

After a full DB reset and re-ingest, co-recall edges are lost. The ingest pipeline extracts entries from session files but does not create edges between entries extracted from the same session. This means graph-augmented recall (#297) has no data to work with until edges accumulate organically through live usage.

## Proposed Solution

During ingest, after extracting entries from a session file, create co-recall edges between all entries extracted from (or already existing in) the same session. Entries from the same conversation are naturally associated - this is an even stronger signal than "recalled together" since they were literally discussed together.

### Implementation

1. After a session file is ingested and entries are stored, collect all entry IDs produced from that session
2. Call `strengthenCoRecallEdges(db, entryIds)` with those IDs
3. This uses the existing edge strengthening logic (weight += 0.1, capped at 1.0, normalized pair order)

### Edge Cases

- Sessions that produce only 0-1 entries: skip (need 2+ for edges)
- Re-ingest of already-ingested sessions: edges get strengthened again (harmless, weight increases slightly)
- Large sessions with many entries: O(n^2) pairs, but capped by the natural session size (rarely more than 20-30 entries per session)

## Why This Matters

- Graph-augmented recall (v0.9.9) depends on edges existing to surface associatively connected entries
- Without ingest-time edges, a fresh install or DB reset starts with an empty graph
- The backfill script (scripts/backfill-edges-v2.ts) proved the approach works with 2,039 edges from timestamp clustering
- Baking it into ingest makes it permanent and automatic

## Acceptance Criteria

- [ ] Ingest creates co-recall edges between entries from the same session
- [ ] `agenr edges` shows edges after a fresh ingest
- [ ] Existing ingest tests still pass
- [ ] At least 2 new tests: edges created for multi-entry session, no edges for single-entry session
- [ ] CHANGELOG entry

Depends on: #297 (graph-augmented recall)

eja-assistant:
## 0.9.10

Bundles three issues into one release:

### #300 - Ingest co-recall edges
- Ingest creates co-recall edges between entries extracted from the same session file, seeding the graph for graph-augmented recall after fresh installs or DB resets
- Backfills edges for already-ingested files on re-ingest without requiring `--force`
- Edge count validation detects and repairs partial backfills

### #263 - In-memory consolidate similarity
- `buildClusters` and `consolidateRulesEntries` now compute cosine similarity in-memory using already-loaded embeddings instead of per-entry SQLite `vector_top_k` queries
- Eliminates O(N) database round-trips from consolidate clustering

### #264 - LLM dedup pre-screening
- Two-band clustering: tight band (>= 0.82) auto-unions as before, loose band (>= 0.65) uses subject-aware auto-union or LLM pre-screening
- `llmDedupCheck` asks the model whether two entries express the same knowledge
- Stats (llmDedupCalls, llmDedupMatches) plumbed through all consolidation phases
- Loose diameter floor for validation of loose-union clusters

### Tests
- 1615 tests passing (112 files)
- New tests: edge creation, backfill, partial backfill, dry-run skip, error resilience, loose-band clustering, LLM dedup mock, rules.ts in-memory similarity

Closes #300, closes #263, closes #264

## Summary by CodeRabbit

* **New Features**
  * Co-recall edges now created for entries from the same session file and backfilled on re-ingest to seed graph-augmented recall.
  * Consolidation supports a loose similarity band with subject-aware auto-union and optional LLM pre-screening for semantic deduplication.

* **Performance Improvements**
  * Consolidation uses in-memory pairwise cosine similarity, removing per-entry DB round-trips.

* **Reports**
  * Consolidation reports now show LLM deduplication call and match counts per phase and in totals.

* **Tests**
  * Added coverage for co-recall edge behavior and LLM dedup scenarios.


jdvmi00:
@coderabbitai review

jdvmi00 (approved):
LGTM - 6 commits, 1615 tests passing, CRabbit findings addressed

jdvmi00 (approved):
LGTM - 6 commits, 1615 tests passing, CRabbit findings addressed

PR #305, opened 2026-02-27
fix: add 15s timeout to LLM dedup pre-screening to prevent hangs

eja-assistant:
Hotfix for consolidate hanging in 0.9.10.

`runSimpleStream` calls in `llmDedupCheck` hang indefinitely when the LLM endpoint is unresponsive. Wraps the call in `Promise.race` with a 15s timeout. On timeout, returns `false` (entries treated as distinct) which is the safe default.

Discovered during first real-world consolidate run after DB reset + full re-ingest with 1215 entries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added 15-second timeout to consolidate clustering operations to prevent hangs from unresponsive endpoints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.9.11 - 2026-02-27
+
+### Fixed
+- Added 15s timeout to LLM dedup pre-screening calls in consolidate clustering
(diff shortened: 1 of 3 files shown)


jdvmi00 (approved):
Hotfix LGTM

PR #306, opened 2026-02-27
fix: register pi-ai API providers before streamSimple calls

eja-assistant:
Root cause fix for consolidate hang in 0.9.10/0.9.11.

**Problem:** `tsup` tree-shook the pi-ai provider registration side-effect import (`register-builtins.js`), so `streamSimple` could not resolve the API backend (`openai-responses`, `anthropic-messages`, etc.). This caused `llmDedupCheck` in consolidate clustering to hang indefinitely (the error was swallowed by the catch block, leaving the process idle).

**Fix:** Explicitly import `registerBuiltInApiProviders` from `@mariozechner/pi-ai` and call it via an `ensureApiProviders()` guard in `runSimpleStream` before any `streamSimple` call that does not use a provided `streamSimpleImpl`.

**Also fixes:** The same latent bug in `merge.ts` which calls `runSimpleStream` without `streamSimpleImpl` - merge would have hung too once clustering stopped blocking it.

1615 tests passing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where consolidation-based LLM dedup/merge functionality was not resolving streaming backends correctly. API providers are now automatically initialized before stream creation to ensure proper operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## 0.9.11 - 2026-02-27
 
 ### Fixed
+- Register pi-ai API providers before streamSimple calls so consolidate
+  LLM dedup and merge can resolve the OpenAI/Anthropic streaming backend.
+  Root cause: tsup tree-shook the side-effect import; now uses explicit
+  registration via ensureApiProviders() guard.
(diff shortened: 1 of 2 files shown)


jdvmi00 (approved):
Root cause fix LGTM

PR #307, opened 2026-02-27
chore: bump to 0.9.12

eja-assistant:
Version bump for npm publish (0.9.11 was already published with timeout-only fix, this includes the root cause provider registration fix).


jdvmi00 (approved):
LGTM

PR #308, opened 2026-02-27
fix: make LLM dedup pre-screening opt-in via --loose-threshold

eja-assistant:
LLM dedup in consolidate clustering caused hundreds of sequential API calls, making consolidate appear to hang. Now only enabled when --loose-threshold is explicitly passed.

## Release Notes v0.9.13

* **New Features**
  * Added `--loose-threshold` command-line flag to enable LLM-based deduplication pre-screening during data consolidation operations

* **Changed**
  * LLM-based deduplication pre-screening during consolidation is now opt-in through the `--loose-threshold` flag. By default, when this flag is not specified, the consolidation process employs subject-aware auto-union for deduplication while avoiding LLM API calls


jdvmi00 (approved):
LGTM

PR #309, opened 2026-02-27
fix: resolve pi-ai dual module registry causing LLM dedup hang

### LINKED ISSUE #314: feat: subsequent-turn auto-recall in OpenClaw plugin
## Summary

Add heuristic-gated memory recall to subsequent turns in the OpenClaw plugin's before_prompt_build hook. Currently, full recall only runs on the first turn of a session. This change adds mid-session recall that fires when the user's message warrants it.

## Problem

The agent has memory tools but doesn't use them reliably mid-session. The plugin already hooks every user message via before_prompt_build but only checks for signals after the first turn. If the user mentions a person, project, or past decision on turn 15, the agent has no automatic recall - it must manually call agenr_recall.

## Solution

- Regex-based message classifier (trivial/normal/complex) determines if recall is warranted
- Sliding window buffer of last 5 user messages builds contextual queries
- Jaccard similarity check prevents redundant recalls on the same topic
- Results deduplicated against session-start recalled entries
- Injected under a distinct `## Recalled context` header
- Session-start recall behavior is completely unchanged

## Key Design Decisions

- Classifier constants are hardcoded (linguistic constants, not user preferences)
- Behavioral thresholds are configurable via midSessionRecall plugin config
- No cooldown - similarity check naturally prevents redundant recalls
- Flat token budgets (no decay with session length)
- ~600ms latency per recall via CLI (acceptable, hidden behind LLM response time)

## Test Plan

- 16-case message classifier test suite
- Query builder, similarity check, and state management tests
- Integration tests for mid-session recall in before_prompt_build

## References

- Research: docs/internal/always-on-memory-research-2026-02-27.md
- Spec: docs/internal/subsequent-turn-recall-spec.md
- Prompt: docs/internal/prompts/prompt-314-subsequent-turn-recall.md

eja-assistant:
## Problem

tsup rewrites the `registerBuiltInApiProviders` import to a deep subpath (`@mariozechner/pi-ai/dist/providers/register-builtins.js`) while `streamSimple` stays imported from `@mariozechner/pi-ai`. In ESM these resolve to separate module instances with separate API registries - providers get registered in one but looked up from another.

This caused every `runSimpleStream` call to fail silently. The LLM dedup check in consolidate clustering hit the 15s timeout and returned false, making the feature useless.

## Fix

- Import `streamSimple` from `@mariozechner/pi-ai/dist/stream.js` which self-registers providers via side-effect import on module load - same registry instance, no separate registration needed
- Remove `ensureApiProviders()` / `registerBuiltInApiProviders` entirely from `src/llm/stream.ts`
- Revert 0.9.13 workaround that made LLM dedup opt-in only

## Verification

Built dist confirmed clean:
- No `registerBuiltInApiProviders` in dist/
- No `register-builtins` deep import in dist/
- `streamSimple` imported from `@mariozechner/pi-ai/dist/stream.js`

Closes #314

## Release Notes v0.9.14

* **Bug Fixes**
  * Fixed a module registry issue affecting streaming functionality and LLM deduplication behavior in bundled deployments.
  * Reverted previous LLM deduplication workaround.

* **New Features**
  * Added new deduplication check utility for API users.


jdvmi00 (approved):
LGTM

PR #312, opened 2026-02-27
perf: batch and parallelize consolidate LLM dedup checks

eja-assistant:
Batches up to 10 dedup pairs per API call with 5 concurrent batches. Reduces a 2473-pair dedup queue from ~60min (sequential, 1 call/pair) to ~2min.

- New `llmDedupCheckBatch` sends multiple pairs in a single tool call
- Concurrent batch groups via `Promise.allSettled`
- Conservative fallback: failed/partial batches default to same=false
- Timer leak fix: `clearTimeout` on both single and batch dedup checks
- Progress logging with ETA based on checked pairs

1626 tests passing.

## Release Notes - Version 0.9.17

* **New Features**
  * Batch deduplication: Improved processing speed by executing multiple deduplication checks concurrently. Typical workloads now complete in ~2 minutes instead of ~60 minutes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.9.17 - 2026-02-27
+
+### Changed
+- Optimized LLM dedup in consolidate clustering: batch up to 10 pairs per API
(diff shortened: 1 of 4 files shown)


jdvmi00 (approved):
LGTM

PR #328, opened 2026-02-28
fix(openclaw): simplify mid-session recall to two-tier classification

### LINKED ISSUE #326: Simplify mid-session recall to two-tier classification
## Problem

The current three-tier classification (trivial/normal/complex) relies on fragile regex lists (EXPLICIT_RECALL, TEMPORAL_PATTERNS) that keep missing edge cases. Example: "What's the status of the exploration agents idea?" classified as `normal` (limit=5) instead of `complex` (limit=8) because the regex matches "status with" but not "status of". One preposition off = missed recall.

Maintaining these regex lists is whack-a-mole. Every new miss requires a new pattern.

## Solution

Collapse to two tiers:

**Tier 1: Trivial (skip recall entirely)**
- Exact matches (ok, sure, thanks, etc.)
- Phrase matches (sounds good, go for it, etc.)
- Short messages (<=8 words) with no entities, no question marks, no temporal markers

**Tier 2: Everything else (fire recall, limit=8)**

### What to change

1. Remove `normal` from `MessageClassification` type - only `trivial` | `recall`
2. Remove `EXPLICIT_RECALL` and `TEMPORAL_PATTERNS` as classification drivers
3. Remove `DEFAULT_MID_SESSION_NORMAL_LIMIT` and `DEFAULT_MID_SESSION_COMPLEX_LIMIT` - replace with single `DEFAULT_MID_SESSION_RECALL_LIMIT = 8`
4. Remove `midSessionRecall.normalLimit` and `midSessionRecall.complexLimit` config - replace with single `midSessionRecall.limit`
5. Simplify `classifyMessage()` to only distinguish trivial vs non-trivial
6. Update all tests

### Files
- `src/openclaw-plugin/mid-session-recall.ts`
- `src/openclaw-plugin/index.ts`
- `src/openclaw-plugin/index.test.ts`

## Rationale

- The trivial gate is robust (structurally simple messages are easy to detect)
- The dedup stash is the real guard against over-injection, not the recall limit
- 8 results for non-trivial messages costs ~800-1200 tokens - low cost vs the cost of a recall miss
- Eliminates ~15 regex patterns from classification maintenance

eja-assistant:
## Summary

Simplifies mid-session recall from three-tier classification (trivial/normal/complex) to two-tier (trivial/recall). All non-trivial messages now use a single recall limit of 8.

## Problem

The three-tier system relied on fragile regex lists (EXPLICIT_RECALL, TEMPORAL_PATTERNS) to distinguish normal from complex. These kept missing edge cases - e.g. "What's the status of the exploration agents idea?" classified as normal (limit=5) because the regex matched "status with" but not "status of".

## Changes

- Collapsed MessageClassification to `trivial | recall`
- Removed normal/complex branching in classifyMessage() - everything past trivial returns `recall`
- Single DEFAULT_MID_SESSION_RECALL_LIMIT = 8
- Config: single `limit` field replaces deprecated `normalLimit`/`complexLimit`
- Updated all tests, added 3 new cases

Net -6 lines.

Closes #326

## Summary by CodeRabbit

* **Improvements**
  * Streamlined mid-session recall with a simplified two-tier classification model.
  * Unified recall limit configuration: all non-trivial messages now use a single limit (default: 8).
  * Legacy configuration fields are now deprecated; use the new single limit field instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 0.9.21 (2026-02-28)
+
+### Improvements
+- OpenClaw plugin: simplified mid-session recall from three-tier
+  classification (trivial/normal/complex) to two-tier (trivial/recall).
+  All non-trivial messages now use a single recall limit of 8, eliminating
+  fragile regex-driven classification that missed edge cases like
+  "What's the status of X?" being classified as normal instead of complex.
+  Config fields normalLimit and complexLimit are deprecated in favor of
+  a single limit field. (#326)
+
 ## 0.9.20 (2026-02-28)
 
 ### Bug Fixes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenr",
-  "version": "0.9.20",
+  "version": "0.9.21",
   "openclaw": {
     "extensions": [
       "dist/openclaw-plugin/index.js"
--- a/src/openclaw-plugin/index.ts
+++ b/src/openclaw-plugin/index.ts
@@ -54,8 +54,7 @@ const SKIP_SESSION_PATTERNS = [":subagent:", ":cron:"];
 const DEFAULT_MAX_SEEN_SESSIONS = 1000;
 const seenSessions = new Map<string, true>();
 const MAX_RECALLED_SESSIONS = 200;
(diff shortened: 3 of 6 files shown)


jdvmi00 (approved):
Clean simplification. LGTM

jdvmi00 (approved):
Clean simplification. LGTM

PR #395, opened 2026-03-02
fix: apply --limit flag to reflection task cluster count (#385)

### LINKED ISSUE #385: fix: --limit flag not applied to reflection task cluster count
## Problem

The `--limit` CLI flag for `agenr maintain` does not cap the number of clusters processed by the reflection task. Running `--only reflection --limit 1` still processes all 106 qualifying clusters.

## Expected

`--limit N` should cap the number of clusters processed (and thus reflections created) to N, consistent with how it caps items in conflicts, consolidation, and retirement tasks.

## Context

Discovered during first production reflection run on v0.9.42. The limit is wired into `TaskExecutionContext` but the reflection task loop in `src/commands/maintain/tasks/reflection.ts` does not check `context.options.limit`.

## Fix

Add a check in the cluster processing loop:
```typescript
if (reflectionsCreated >= (context.options.limit ?? Infinity)) break;
```

Low priority - the first full-brain run needed to process everything anyway.

eja-assistant:
## Summary

The `--limit` CLI flag was not applied to the reflection task. Running `--only reflection --limit 1` would still process all qualifying clusters.

## Fix

Added limit check at the top of the cluster processing loop in `runReflectionTask`. Uses `reflectionsCreated` as the counter (not `clustersProcessed`) since clusters can be processed without producing a reflection (verification failure, LLM error, etc.).

## Files Changed
- `src/commands/maintain/tasks/reflection.ts` - 7 lines added
- `tests/commands/maintain/tasks/reflection.test.ts` - 1 new test

## Tests
2,060 passing (140 files).

Closes #385


jdvmi00 (approved):
LGTM - 2060 tests passing, clean 7-line fix

PR #398, opened 2026-03-02
feat: suppress demoted source entries when reflection is in recall results (#389)

### LINKED ISSUE #389: Suppress demoted source entries when their reflection is in recall results
## Problem

When a reflection and its demoted source both match a recall query, both appear in results. Example: 'agenr PR structuring preference' shows as both a reflection (imp 8, position 3) and a demoted preference (imp 6, position 4). This wastes recall budget with redundant information.

## Proposed Fix

During recall result assembly, if a reflection entry is in the result set, check `entry_sources` for its source entries and exclude any that also appear in the results. The reflection is by definition a synthesis of those sources - showing both is redundant.

## Considerations

- Only suppress when the reflection is actually present in results
- If a source entry matches a very different query aspect than the reflection, it might still be valuable (edge case)
- Could be implemented as a post-scoring filter step

eja-assistant:
## Summary

When a reflection and its demoted source entries both match a recall query, the sources are now filtered out since the reflection is a synthesis of those sources. Results are backfilled from the scored pool to maintain the requested limit.

## Changes

- New `src/db/recall/suppress.ts` with `suppressReflectedSources()`
- Integrated into recall pipeline after scoring/limiting
- Backfill logic when suppressed results drop below limit
- Chunked queries to entry_sources for reflection source lookup

## Tests
2,067 passing (141 files). 4 new tests in `tests/db/recall-suppress.test.ts`.

Closes #389

## Summary by CodeRabbit

* **New Features**
  * Recall results now suppress demoted source entries when their reflection appears in results, with automatic backfilling to maintain the requested result limit.

* **Chores**
  * Version bumped to 0.9.44.


jdvmi00 (approved):
LGTM

PR #413, opened 2026-03-02
fix: rewrite verification prompt, parallelize reflections, bump store timeout (#409, #411, #412)

eja-assistant:
### Verification Prompt Rewrite (#411)
- Reframes verification from fact-checking to synthesis quality review
- Checks accuracy (misrepresentation) separately from theme coverage
- Omissions are expected and no longer penalized
- Lowers default coverage threshold from 0.7 to 0.5
- Adds parse error retry with JSON-only nudge and permissive fallback
- Expected: verify failures drop from 27% to ~2%

### Reflection Parallelization (#409)
- Extracts cluster processing into standalone `processCluster` function
- Processes clusters in configurable parallel batches (default: 5)
- Adds `maintain.reflectionConcurrency` config knob
- Adds per-batch progress logging with ETA (#404)
- Adds SQLITE_BUSY retry/backoff for concurrent write safety
- Expected: runtime drops from ~29 min to ~8-12 min

### Store Timeout Bump (#412)
- Bumps TOOL_TIMEOUT_MS from 10s to 30s
- Fixes agenr_store timeouts with larger brains (2,600+ entries)
- Updates timeout test to match new value

## Test Results
2,119 tests passing across 143 files.

Closes #409, #411, #412. Also addresses #404 (progress counter).

## Release Notes – Version 0.9.46

* **New Features**
  * Added configurable `reflectionConcurrency` setting (default: 5) for parallel reflection cluster processing.

* **Improvements**
  * Enhanced reflection verification to evaluate synthesis quality instead of completeness.
  * Lowered reflection coverage threshold from 0.7 to 0.5.
  * Added automatic retry mechanism for reflection parsing errors.
  * Improved SQLite contention handling with backoff logic.
  * Extended OpenClaw plugin store timeout to 30 seconds.
  * Added per-batch progress logging and checkpoint resumption support.

PR #415, opened 2026-03-02
fix: add reflection type demotion in recall scoring

eja-assistant:
## Problem

After running the reflection pipeline, 80 new reflections dominate recall results because they have high importance (8), freshness bonus (1.5x), and broad vector similarity. Specific raw facts get buried - e.g., asking "what model does extraction use" returns architectural reflections instead of the actual answer (gpt-4.1-nano).

## Solution

Add a configurable `reflectionFactor` multiplier (default 0.85) applied to reflection-type entries during recall scoring, following the same pattern as the existing `todoPenalty`. This 15% demotion lets highly relevant raw facts compete with broadly relevant reflections.

- New constant: `REFLECTION_DEMOTION = 0.85`
- New config knob: `scoring.reflectionDemotion` (range 0-1)
- Exposed in scores breakdown as `reflectionFactor`
- Config normalization with validation

## Tests

6 new tests covering demotion application, non-reflection passthrough, config override, invalid config fallback, scores object exposure, and config normalization. 2,127 total tests passing.

Closes the reflection crowding concern from PR #413 review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recall scoring to prevent reflection entries from dominating results through a new reflection demotion factor (default 0.85x multiplier).

* **Documentation**
  * Added `scoring.reflectionDemotion` configuration field with usage examples and range specifications (0-1).


jdvmi00 (approved):
LGTM - clean scoring multiplier, follows todoPenalty pattern. 2,127 tests passing.

Issue #388, opened 2026-03-02
Reflections should inherit/synthesize tags from source entries


eja-assistant:
## Problem

All 83 reflection entries have zero tags (`tags: none` in recall output). This limits filtering and discoverability.

## Current Behavior

The reflection pipeline creates entries with empty tags. Source entries often have useful tags (e.g., `agenr`, `architecture`, `refactoring`) that could inform the reflection's tags.

## Proposed Fix

During reflection creation, collect tags from all source entries, pick the top N most common (e.g., top 5), and assign them to the reflection entry. Alternatively, ask the LLM to suggest tags as part of the synthesis prompt.

## Impact

Low effort, high value for recall filtering and context injection.

PR #421, opened 2026-03-02
fix(recall): cap freshness boost at 1.0 for reflection entries (#410)

### LINKED ISSUE #410: Demote reflections in recall scoring to preserve raw entry visibility
## Problem

After a full reflection pass, recall results are 100% reflections. Raw entries are completely crowded out because reflections are fresh (freshness bonus 1.25-1.5x), high importance (8), and get cluster boost.

A query like 'what was the exact error message' should surface the raw entry, not a paraphrased reflection.

## Options

1. Cap freshness bonus for reflections at 1.0 (no freshness boost for type=reflection)
2. Apply a type factor: reflections get 0.85x, raw entries get 1.0x
3. Both

## Expected Impact

- Restores raw entry visibility in recall
- Improves precision for detail-specific queries
- Reflections still rank well due to importance and cluster boost, just not exclusively

eja-assistant:
## Summary

Reflection entries are always freshly created (just synthesized during maintain), so they were receiving a 1.1x–1.5x freshness bonus on top of their high importance (8). This caused reflections to dominate recall results even with the existing 0.85x reflectionFactor demotion.

## Change

Added a single guard at the top of `freshnessBoost` in `src/db/recall/score.ts`:

```typescript
if (entry.type === "reflection") return 1.0;
```

Reflections are synthesized summaries — freshness is not meaningful for them. Raw entries now have a fair shot at surfacing alongside reflections.

## Tests

3 new tests in `tests/db/recall.test.ts`:
- reflection < 1 hour old returns 1.0 (would have been 1.5)
- reflection < 6 hours old returns 1.0 (would have been 1.25)
- non-reflection entries still receive the freshness boost as before

Ships as part of v0.9.48 alongside #388.

Closes #410

## Summary by CodeRabbit

* **Bug Fixes**
  * Capped the freshness bonus at 1.0 for reflection entries to improve recall result ranking accuracy.

* **Tests**
  * Added unit tests for freshness scoring behavior across entry types with specific age thresholds.


jdvmi00 (approved):
LGTM

jdvmi00 (approved):
LGTM

PR #482, opened 2026-03-04
feat: pre-store embedding dedup layer (#474)

### LINKED ISSUE #474: feat: extraction-time dedup (similarity check before insert)
## Problem
The extraction pipeline has no awareness of what's already stored. Every session transcript produces ~200 new entries/day regardless of overlap with existing knowledge. This is the root cause of unbounded DB growth.

## Solution
Before inserting each extracted entry:
1. Compute its embedding
2. Check against top-K existing entries (K=5-10) via ANN index
3. If similarity > 0.80, skip or merge inline rather than insert as new

**Expected impact**: 40-60% reduction in new entries (~80-120/day instead of ~200)

**Cost**: One embedding lookup per candidate (~10ms). Negligible vs. downstream consolidation cost.

## Architectural evolution
This could grow into an extract-then-merge single pass:
```
Session transcript
  → LLM extracts candidate entries
  → Embedding similarity check against existing DB
  → For each candidate:
      if similar entry exists: merge inline (update existing)
      if genuinely new: insert
```
Background consolidation still runs for historical cleanup but at much lower pressure.

## Related
- Brain health audit: `docs/internal/brain-health-report-2026-03-04.md`
- #471 (dynamic consolidation caps)

jdvmi00:
## Summary

Adds a pre-store embedding dedup layer that catches near-duplicate knowledge entries **before** they hit the store pipeline, reducing DB bloat and consolidation pressure.

Closes #474

## How It Works

1. After LLM extraction + trigram dedup, batch-embed all candidate entries (single API call)
2. Check each against DB via `findSimilar` (K=5, DiskANN vector index)
3. If cosine similarity >= 0.90 **and same type** -> skip
4. Events are excluded from dedup entirely (distinct occurrences)
5. Within-batch pairwise cosine similarity catches same-batch dupes (keeps longer entry)
6. Pre-computed embeddings are returned and threaded through the write queue -> store pipeline via `EmbeddingCache`, eliminating double embedding API costs

## Key Design Decisions

- **Threshold 0.90** (not 0.80) -- avoids a dead zone where entries that the store's nuanced logic (reinforce at 0.88, LLM merge at 0.72) would handle get silently dropped
- **sameType gate** -- a fact and decision at 0.91 are genuinely different entries
- **Event exclusion** -- "deployed v2.1" on Monday and Tuesday are distinct occurrences
- **No runExclusive** -- dedup is read-only, no write lock needed during embedding API calls
- **Embedding cache passthrough** -- verified end-to-end (8 links), zero re-embedding for survivors

## Files Changed

- `src/commands/ingest/pre-store-dedup.ts` -- new dedup module
- `src/commands/ingest/target-processor.ts` -- pipeline integration
- `src/db/store/index.ts` -- `embeddingCache` option on `StoreEntriesOptions`
- `src/ingest/write-queue.ts` -- cache threading through write queue
- `tests/ingest/pre-store-dedup.test.ts` -- 11 tests

## Review History

- 4-agent review swarm on initial implementation (correctness, performance, tests, architecture)
- Architecture reviewer recommended raising threshold from 0.80 to 0.90 and skip-only at that level
- Codex implemented fixes from review prompt
- 3-agent re-review swarm on Codex's rewrite (checklist, correctness, performance)
- Final fixes: within-batch sameType gate, keep-longer logic, version revert

## Summary by CodeRabbit

* **New Features**
  * Added automatic deduplication during ingest to filter similar entries before storage with a conservative default setting.
  * Type-safe deduplication ensures only entries of the same type are compared.
  * Event-type entries are now exempt from deduplication filtering.
  * Expanded metrics tracking for ingest operations with additional counters for storage, skipped, and reinforced entries.

* **Performance**
  * Optimized embedding computation by reusing cached embeddings across the ingest pipeline.


eja-assistant (approved):

